PR #4, opened 2017-07-12
Add Material-UI

oliviertassinari:
great project


ai:
Awesome! :)

oliviertassinari:
capture d ecran 2017-07-13 a 18 23 22

ai:
For such big project — yeap. But it is always good idea to make it smaller :D.

oliviertassinari:
We have done some effort in that direction lately. I don't want to see any regression :). Thanks for sharing this project.

PR #6, opened 2017-07-14
fix support glob patterns test

pgrekovich:
If you using `fixture()` function, you will get full file name, and this test didn't work.

### DIFF
--- a/test/cli.test.js
+++ b/test/cli.test.js
@@ -110,7 +110,7 @@ it('shows limit', () => {
 })
 
 it('supports glob patterns', () => {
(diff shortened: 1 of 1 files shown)


ai:
Glob pattern is supported only for new `sizeLimit` section API.

`size-limit path/to/js` is depreacted in some sort.

Is it OK? Or you need old CLI API?

pgrekovich:
No, i don't need old CLI, sorry. 
But, can it make sense to remove this test?

ai:
Glob is using inside `test/fixtures/glob/package.json`

PR #17, opened 2017-09-05
Losslessly minify logo

subzey:
Optimizations svgo cannot do: Remove useless fill-rules, rearrange and merge black paths, simplify white path.


ai:
Wow, thanks

PR #25, opened 2017-09-29
Error with Webpack Bundle Analyzer start

Tereshin:
I had problem with Webpack Bundle Analyzer start after ```npx size-limit --why``` command.
In ```cli.js``` file I described it.


ai:
You forgot to run tests before PR ;)

PR #31, opened 2017-10-31
Support more involved custom webpack

SivanMehta:
This fixes a bug where the dependencies of the custom webpack configuration were not loaded correctly. For example, at the top of my `webpack.config.js` in a private project I have the following:

```js
const ExtractTextPlugin = require('extract-text-webpack-plugin');
const importer = require('sass-import-modules').importer;
```

Because these are not dependencies of `size-limit` (and shouldn't be), they were not being loaded correctly **_relative to the module_**. This PR fixes that. Adjusted results now that this works

```
➜  uxcore2 git:(size) npm run size

  Package size: 1.55 MB
  With given webpack configuration
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -126,6 +126,15 @@ function runWebpack (config, opts) {
   })
 }
 
+function extractSize (stat, opts) {
+  let name = `${ stat.compilation.outputOptions.filename }`
+  name += opts.config ? '' : '.gz'
+  const assets = stat.toJson().assets
+  const size = assets.find(i => i.name === name).size
+  console.log(size)
+  return size
+}
+
 /**
  * Return size of project files with all dependencies and after UglifyJS
  * and gzip.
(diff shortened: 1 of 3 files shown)


test/cli.test.js
```diff
@@ -244,13 +244,6 @@ it('throws on --why with disabled webpack', () => {
   })
 })
 
-it('uses custom webpack', () => {
-  return run([], { cwd: fixture('config') }).then(result => {
-    expect(result.out).toContain('Package size: 2.2 KB')
-    expect(result.code).toEqual(0)
-  })
-})
-
```
SivanMehta: This tests was removed as it was a duplicate of the below test

ai: We need both tests since one test API and second test CLI (integration test).

SivanMehta: Good point, I will put it back.

test/fixtures/config/webpack.config.js
```diff
@@ -5,6 +5,7 @@ const path = require('path')
 module.exports = {
   entry: path.join(__dirname, 'index.js'),
   output: {
+    path: __dirname,
```
ai: `output.path` is requires for all webpack configs?

index.js
```diff
@@ -116,6 +126,13 @@ function runWebpack (config, opts) {
   })
 }
 
+function extractSize (stat, opts) {
+  let name = `${ stat.compilation.outputOptions.filename }`
```
ai: Do we need quotes here? :D

index.js
```diff
@@ -116,6 +126,13 @@ function runWebpack (config, opts) {
...
+  name += opts.config ? '' : '.gz'
```
ai: Will it used gzipped size in custom config?

SivanMehta: It will use the output file name specified in the desired webpack configuration. So if you specify it, it will pick it up.

ai: What users need to do to show gzipped size?

SivanMehta:
@ai addressed your feedback

ai:
Looks good :)

ai:
Released in 0.12.1

PR #32, opened 2017-11-05
Added config file support

borovik96:
Do I need to write more tests?

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,22 @@ If you don't have a continuous integration service running, don’t forget
 to add one — start with [Travis CI](https://github.com/dwyl/learn-travis).
 
 ## Config
-You can use config file `.size-limit` instead of `package.json` setting:
+Add `size-limit` section to `package.json` and `size` script:
+
+```json
+  "size-limit": [
+    {
+      "path": "index.js",
+      "limit": "9 KB"
+    }
+  ],
+  "scripts": {
+    "size": "size-limit",
+    "test": "jest && eslint ."
+  }
+```
+
+Also you can use config file `.size-limit` instead of `package.json` setting:
 
 ```json
   [
--- a/README.md
+++ b/README.md
@@ -81,6 +81,20 @@ Add `size-limit` section to `package.json` and `size` script:
   }
 ```
 
+or add `.size-limit` config in your root directory:
+
+```json
+  [
+    {
+      "path": "index.js",
+      "limit": "9 KB"
+    }
+  ]
+```
+
+Config from `package.json` will rewrite `.size-limit` config.
+
+
 Here's how you can get the size for your current project:
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -81,20 +81,6 @@ Add `size-limit` section to `package.json` and `size` script:
   }
 ```
 
-or add `.size-limit` config in your root directory:
-
-```json
-  [
-    {
-      "path": "index.js",
-      "limit": "9 KB"
-    }
-  ]
-```
-
-Config from `package.json` will rewrite `.size-limit` config.
-
-
 Here's how you can get the size for your current project:
 
 ```sh
@@ -138,6 +124,20 @@ Add the `size` script to your test suite:
 
 If you don't have a continuous integration service running, don’t forget
 to add one — start with [Travis CI](https://github.com/dwyl/learn-travis).
+
+## Config
+You can use config file `.size-limit` instead of `package.json` setting:
+
+```json
+  [
+    {
+      "path": "index.js",
+      "limit": "9 KB"
+    }
+  ]
+```
+
+**NOTE**: setting in `package.json` will rewrite `.size-limit` config.
 
 ## Applications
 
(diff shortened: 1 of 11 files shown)


ai:
1. Do we need `readPkg`? `cosmiconfig` can load config from `package.json` too.
2. Users don’t like to have choice. Maybe it is better to tell about `.size-limit` config in separated section (or Configs section). So in `Usage` section we will still be very focused on the most popular solutions.

borovik96:
As I understand, with cosmiconfig we can't handle both names at **package.json** 'size-limit' and 'sizeLimit'.

ai:
I am OK with removing legacy `sizeLimit`. Anyway, we showed a warning during few minor versions. Also because we are in `0.x` versions, we can break public API in minor versions ;).

borovik96:
Should I check existing of `package.json`?

ai:
If we have `.size-limit`, we don’t need `package.json`.

ai:
1. If we want good docs, we should always think that user didn’t read whole docs. Right now `Config` section tells about only `.size-limit` (user may miss Usage section and don't read accurately). So I think it will be better to explicit say that there is a 2 ways. And show both ways together.
2. We need to update error messages as well.

borovik96:
Tests are failed cause there is spelling checking, but I  didn't change thet part of `README`

ai:
Ouh. It is Yaspeller problem. I will fix it.

borovik96:
Appveyor test failed on `index.test.js`, I have no idea about this trouble, I didn't touch this stuff)

borovik96:
About error messages, what do I need to fix? Can I add a link to the Config section at GitHub?

ai:
Check out origin error messages. You will see that error messages were clear and tell exactly what and where should be changed.

> settings for `"size-limit"

Why `"size-limit"` in quotes?

> Add it according to Size Limit docs

Add where?

> Can not parse `"size-limit"` config

What config and what it is exacty a error?

borovik96:
> Why "size-limit" in quotes?
Because it was so before

ai:
> Because it was so before

Before it was about `package.json` section. This is a reason why it was in quotes. In your case, you don’t need to use quotes. `Size Limit settings"` is correct.

borovik96:
>you don’t need to use quotes

Ok, I will fix it. 

Do you know what may be wrong with AppVeyor CI, how can I fix it?

ai:
Note, that `"size-limit"` quotes is not the only problem.

borovik96:
I have fixed error messages, now the parsing error contains the path to the config and explanation of the error.

README.md
```diff
@@ -125,6 +125,35 @@ Add the `size` script to your test suite:
 If you don't have a continuous integration service running, don’t forget
 to add one — start with [Travis CI](https://github.com/dwyl/learn-travis).
 
+## Config
+Add `size-limit` section to `package.json` and `size` script:
```
ai: No newline after the title.

It is better to explicit say about 2 types of config.

cli.js
```diff
@@ -103,62 +131,43 @@ if (ciJobNumber() !== 1) {
...
-          ignore: result.pkg.peerDependencies,
-          limit: limit.limit,
-          full: files.map(i => path.join(cwd, i)),
-          files
-        }
-      })
-    }))
+    return limits
+  }).then(limits => {
+    return readPkg().then(packageJson => {
```
ai: Why do we need `readPkg` here?

cli.js
```diff
@@ -103,62 +131,43 @@ if (ciJobNumber() !== 1) {
...
+      return Promise.all(limits.map(limit => {
+        const cwd = path.dirname(packageJson.path)
```
ai: For `.size-limit` config you need to use its directory.

test/cli.test.js
```diff
@@ -102,6 +93,27 @@ it('shows size-limit content error', () => {
...
+    '\n')
+    expect(result.code).toEqual(0)
+  })
+})
+
+it('Wrong ".size-limit" file config', () => {
+  return run([], { cwd: fixture('wrong-file-config') }).then(result => {
+    expect(result.out).toContain(
+      ' ERROR  Can not parse Size Limit config: \n' +
+      'missed comma between flow collection entries'
```
ai: All other error messages have indent on next lines.

```
 ERROR  Can not parse Size Limit config:
        missed comma between flow collection entries
```

cli.js
borovik96: Because below we have
```javascript
return {
            webpack: limit.webpack !== false,
            bundle: packageJson.pkg.name,
            config: limit.config,
            ignore: packageJson.pkg.peerDependencies,
            limit: limit.limit,
            full: files.map(i => path.join(cwd, i)),
            files
          }
```

README.md
borovik96: Is it good enough?

image

cli.js
borovik96: Will fix it

test/cli.test.js
borovik96: Will fix it

cli.js
ai: Ouh :(.

README.md
ai: Yeap, looks good.

ai:
Awesome. Thanks. I will accept and release it today evening.

borovik96:
Thanks for your tips!

ai:
Released in 0.13

PR #34, opened 2017-11-13
Clarify installation instructions

### LINKED ISSUE #33: Clarify path option
Which version should I specify in the `path` option: source code or a compiled one (the same I have in the `main` field of package.json)?

sapegin:
1. Fix flow: first install, then add path/script, then check current size, then add limit, then add to CI.
2. Explain what to should be in the `path` option.
3. Don’t use npx, we already have a local package by this time.
4. Improve consistency a bit.

Closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,7 @@ Add `size-limit` section to `package.json` and `size` script:
 ```diff json
 + "size-limit": [
 +   {
-+     "path": "index.js",
-+     "limit": "9 KB"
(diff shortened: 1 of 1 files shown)


ai:
Why you decide to comment `npx`

sapegin:
> Don’t use npx, we already have a local package by this time.

Because would confuse the user: why did he install the local package and how has to use npx (which is still very new and not many people are familiar with).

PR #46, opened 2018-03-02
Remove unnecessary compression

TrySound:

PR #47, opened 2018-03-02
Add correct latest stable version

TrySound:
Some time ago `stable` was deprecated in favor of `node` identifier

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 cache: yarn
 node_js:
-  - stable
+  - node
   - "8"
   - "6"


ai:
Where I can read more about it (seems like I need to change it in all my project :(, maybe there is a change to rename it back?)

TrySound:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions

ai:
Maybe these docs are old. It is big changes there is be an issue or discussion in NVM.

TrySound:
Seems like stable identifier works in the same way. They just probably didn't remove it. Node version is the same. There's some problem with tests. They are not passed on my mac too. But for different reason.

TrySound:
In previous PR tests are failed too
https://github.com/ai/size-limit/pull/46

ai:
Yeap, it is some strange conflict between Webpack 4 and Travis CI :(

Anyway, what is a reason to replace `stable` to `node`? First seems like more clear and declarative for me.

TrySound:
It doesn't exist in docs. And may be removed anytime.

ai:
It was not an official doc (Travis CI uses `nvm`). But I found good [proof link](https://github.com/creationix/nvm#usage) about depreciation.

> stable: this alias is deprecated, and only truly applies to node v0.12 and earlier. Currently, this is an alias for node.

And there we have a reason. `stable` was part of `stable`/`unstable` pair for <= 0.12 development process (where 0.11, 0.9, 0.7 were unstable and 0.12, 0.10, 0.8 was stable).

PR #49, opened 2018-03-03
Remove duplicated uglify and define plugin

TrySound:
Current results are a kind of lie since uglify is applied twice. Webpack by default adds uglify and define plugin with `production` string


ai:
Released in 0.16.

PR #50, opened 2018-03-03
Return from getSize both bundle and gzip results

TrySound:
In my opinion both bundle and gzip results are important since there is also a cost of javascript parsing. This will let to bundle the code only once and have more pretty result. Currently I got this
```
  esm/index.js
  Package size: 29.8 KB

  esm/index.js
  Package size: 9.34 KB
  Size limit:   9.4 KB
```
In the next PR I'm gonna improve render in cli

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -10,27 +10,21 @@ function trim (num) {
   return Math.floor(num / 10) * 10
 }
 
-it('returns 0 for minified empty project', () => {
-  return getSize(fixture('unlimit/empty'), { gzip: false }).then(size => {
-    expect(size).toEqual(0)
-  })
-})
-
-it('returns 0 for gziped empty project', () => {
+it('returns 0 for bundle and gzip empty project', () => {
   return getSize(fixture('unlimit/empty')).then(size => {
-    expect(size).toEqual(0)
+    expect(size).toEqual({ bundle: 0, gzip: 0 })
   })
 })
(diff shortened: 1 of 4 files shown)


TrySound:
Also this is breaking change.

test/index.test.js
```diff
@@ -42,66 +36,66 @@ it('returns error', () => {
...
+    expect(size).toEqual({ bundle: 222, gzip: 97 })
   })
 })
 
 it('disables webpack on request', () => {
   return getSize([
     fixture('bad/index'), fixture('es2016/index')
   ], { webpack: false }).then(size => {
-    expect(size).toEqual(122)
+    expect(size).toEqual({ bundle: 82, gzip: 122 })
```
TrySound: Found some weird thing. Gzipped result is bigger.

ai: Try to build project manually and check output.

TrySound: Yeah, I already tried. Seems like some dictionaries spends more than source.

ai:
Maybe we should rename `bundle` to `parsed` or `minified`? I think the name is not clear (`gzip` is about bundle as well).

TrySound:
What does `parsed` mean?

ai:
`parsed` is a code used in Webpack Bundle Analyzer for the same concept (size of minified JS).

PR #52, opened 2018-03-04
Refactor render and fix array path output

TrySound:
- moved size renderer into pure function
- fixed `index1.js,index2.js` size name because of implicit casting array to string (types are really good here)
- removed a few unused props (and here)
- merged a few `then` chains because async operations are not required between them
- prevented parsing number with `bytes.parse` (and here)


ai:
Released in 0.16.

Issue #19, opened 2017-09-11
Size-limit exits with non-zero if sizes are equal to reported


okonet:
Steps to reproduce:

1. Create a config and run `size-limit` to see the report
1. Update the config with the _exact same number_ from the report
1. Run `size-limit` again and see it failing with

```
  Package has exceeded the size limit
  Package size: 1.8 KB
  Size limit:   1.8 KB
  With all dependencies, minified and gzipped
```

ai:
Seems like we need to show bytes here to explain reason.

ai:
But this issue will be low priority since we recommend to add extra bytes for limits. In you case, most of PR will fall build step. Even if you add one single variable.

okonet:
Makes sense but still very confusing for the first experience. Probably some error boundaries should be backed into the library?

ai:
Can’t reproduce :(. Can you share your project?

okonet:
Been trying it here: https://github.com/okonet/attr-accept/pull/11 and ended up increasing the limit.

ai:
Done c6e45a3

ai:
I will release it with messages fixes.

okonet:
Thanks!

I'm still a bit concerned about the DX here. Now it reports size in bytes, but we specify it in KBs or even MBs :trollface: which will lead to an additional mental overhead for developers. I see at least two ways how this can be improved:

1. Do not show the actual size, instead show the difference.
1. Additionally display pretty formatted sizes 

Something like:

```
Limit exceeded by 142 Bytes!
Actual size: 16343 B (1.6 KB)
Size Limit: 15634 B (1.6 KB)
```

ai:
I like the idea to show the difference! :)

PR #54, opened 2018-03-05
Shows bytes if value is same in different units test

TrySound:
This test was removed. But test checks important part of functionality.

### DIFF
--- a/test/cli.test.js
+++ b/test/cli.test.js
@@ -249,6 +249,14 @@ it('uses different units', () => {
   })
 })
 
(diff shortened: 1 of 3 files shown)


ai:
Nope. This test didn’t reproduce the previous test. The idea of the previous test was:

1. You write `limit: 10KB` (which is 10240 B).
2. Your package size is 10250 B. When you convert it to KB it has the same 10KB.
3. You can’t understand why 10 KB is not passed by 10 KB limit.

Here is original issue beside this test:
https://github.com/ai/size-limit/issues/19

PR #55, opened 2018-03-05
Render both parsed and gzip sizes

TrySound:
Here's new size output

screen shot 2018-03-05 at 7 36 21 pm
screen shot 2018-03-05 at 7 37 08 pm
screen shot 2018-03-05 at 7 37 57 pm

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -136,39 +136,65 @@ function capitalize (str) {
 
 function renderSize (item, i, array) {
   const rows = []
-  const passed = item.limit && item.size <= item.limit
-  const failed = item.limit && item.limit < item.size
-  const unlimited = !item.limit
+  const limit = item.limit
+  const parsed = item.size.parsed
+  const gzip = item.size.gzip
+  const size = typeof gzip === 'number' ? gzip : parsed
+  const passed = limit && size <= limit
+  const failed = limit && limit < size
+  const unlimited = !limit
 
   if (array.length > 1 && item.name) {
     rows.push(item.name)
   }
 
-  let limitString = formatBytes(item.limit)
-  let sizeString = formatBytes(item.size)
+  let limitString = formatBytes(limit)
+  let parsedString = formatBytes(parsed)
+  let gzipString = typeof gzip === 'number' ? formatBytes(gzip) : null
+  const sizeString = gzipString || parsedString
 
   if (passed) {
-    rows.push(
-      `Package size: ${ chalk.bold(chalk.green(sizeString)) }`,
-      `Size limit:   ${ chalk.bold(limitString) }`
-    )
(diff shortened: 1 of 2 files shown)


ai:
You should write a better description.

What is a reason for it?

TrySound:
Gzip is good for loading. But the cost of javascript parsing make sense too. It would be good to have both values in the same place with nicer labels without double bundling.

ai:
But it will confuse users what size is used for limit check. Also, this project goal is to prevent bloating by working in Travis CI.

Most of users will not see parsed size. And most of size bloating issue will bloat both gzip and parsed sizes.

So this UI changes for my point of view will not improve tool.

TrySound:
To distinct target I painted parsed as grey. Probably we can add this feature to unlimited mode? It doesn't have limits and works only as info tool. However I still think this is a good use case when you work on size improvements.

ai:
We will have the same problem in unlimited mode. This mode is used to detect limit. But if unlimited mode will report two different sizes, it will confuse users (what size I should use in the limit).

> However I still think this is a good use case when you work on size improvements.

In this case, you can have two limits with `gzip: false` and without.

TrySound:
Can't focus help?
screen shot 2018-03-05 at 8 13 00 pm

ai:
Nope. It is still harder to explain in docs and harder for users.

ai:
Image yourself going to shop:

> What is a price for milk?
> 10

> What is a price for milk?
> 10 and 20
> What? Why you give me 2 prices? What is a difference between them? What I should use?

TrySound:
When I go to the shop for milk I actually see two prices
> Price: $10
> $7 is the price with our bonus card

ai:
Yeap, any many users complain about it, because they made a mistake ;)

ai:
> you can have two limits with gzip: false and without.

Will this solution cover your use case?

TrySound:
I mentioned that this requires to write more descriptive names and spend more time to bundle twice.
I just think size-limit can give wow effect for other developers when somebody tweaks their babel config. Also it could be like debugging tool when you are looking for difference between runs.

TrySound:
What do you think about the second cli? Something like `size-info`, which will skip gzip and limit options and just show info for everything.

ai:
Second CLI is not compatible with `npx`. But I am OK with extra option like `size-limit --info`.

ai:
Maybe `--sizes` option is better. But how it will work together with `gzip: true/false`. Maybe we need to change it too.

TrySound:
I think gzip false may be skipped. What else can we show?

ai:
What do you mean by “skipped”?

PR #56, opened 2018-03-06
Ignore peer dependencies with multiple entries in why mode

TrySound:


ai:
> No description provided

:man_facepalming: 

Can you explain better what you try to fix there and why?

TrySound:
`why` mode has condition with one size-limit entry in package.json and more than one entries. Currently `ignore` is not passed in the second case even if peerDependencies are present, so when I try to run `why` with esm and cjs entries I get peer dependency in the bundle.

TrySound:
What do you think?

ai:
Oops, I forgot about it =^_^=

ai:
Released in 0.16.2.

PR #60, opened 2018-05-12
chore: clone last 5 commits

DanielRuf:


ai:
Can you tell me more about this setting?

DanielRuf:
Sure. By default Travis CI clones the last 50 commits. Cloning less means faster checkout and clone times which results in slightly faster builds.

See https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth for more information.

ai:
Wow 😍. Should we clone only last commit instead of 5?

DanielRuf:
Also we have to reinstall webpack as just upath 1.0.5 is compatible with the latest Node.js version.

DanielRuf:
> Wow 😍. Should we clone only last commit instead of 5?

> Please note that if you use a depth of 1 and have a queue of jobs, Travis CI won’t build commits that are in the queue when you push a new commit.

Not sure, most projects now use `5`, just very few use `1` which can increase waiting times in the build queue.

PR #61, opened 2018-05-12
chore: reinstall webpack to update upath dependency for Node.js 10 su…

DanielRuf:
…pport

Also the pre-commit hook fails on Node.js 10 because lint-staged wants Node.js 6+ (I guess lint-staged is also restricted to <= 9).


DanielRuf:
`lint-staged` uses `please-upgrade-node` which had a wrong version check.
See https://github.com/typicode/please-upgrade-node/issues/11

PR #63, opened 2018-05-13
Allow the usage of dynamic configuration

oliviertassinari:
Next.js 6 now generates a vendor bundle with a dynamic filename. I'm looking for a way to handle this change. If you are happy with this change, I can update the documentation.
https://github.com/davidtheclark/cosmiconfig#searchplaces

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -176,7 +176,8 @@ function getConfig () {
   const explorer = cosmiconfig('size-limit', {
     searchPlaces: [
       'package.json',
-      '.size-limit'
+      '.size-limit',
+      'size-limit.config.js'
     ]
   })
   return explorer


ai:
How vendor bundle generation related with this changes?

oliviertassinari:
My entry point is dynamic. I can't have a static configuration file.

ai:
Yeap. And why you need to change config name to fix it? Why `.size-limit.config.js` is better than `.size-limit`.

oliviertassinari:
I'm assuming `.size-limit` uses a json loader, while I need a JavaScript loader.

oliviertassinari:
https://github.com/davidtheclark/cosmiconfig/blob/master/README.md#loaders

ai:
It uses YAML loader to be specific.

Why do you need JS loader?

ai:
Get it. Will you generate config on the flight?

oliviertassinari:
> Will you generate config on the flight?

I will. Of course, I could always use the JS API. I'm assuming it's simpler this way.

ai:
I will rename it to `.size-limit.js`

oliviertassinari:
@ai Alright 👍

ai:
Released in 0.18

Issue #68, opened 2018-08-10
`ignore` option breaks Webpack builds


jaydenseric:
If you use the `ignore` option for a dependency that is imported in the file you are testing the size of, you get a Webpack build error like this:

<img width="605" alt="screen shot 2018-08-10 at 5 13 26 pm" src="https://user-images.githubusercontent.com/1754873/43943967-bcbb688c-9cc0-11e8-9f85-b0d0677da025.png">

Also, without using the `ignore` option, if your dependency is a peerDependency and devDependency, but not a dependency, the same error happens. I assume the buggy mechanism for the `ignore` option is being repurposed to exclude peerDependency sizes from the report?

jaydenseric:
Maybe a better approach is using Webpack `externals` config option, with regex?

> Every dependency that matches the given regular expression will be excluded from the output bundles.
> — https://webpack.js.org/configuration/externals/#regex

Also, see [this comment](https://github.com/webpack/webpack/issues/2858#issuecomment-240338772).

ai:
@jaydenseric hm. interesting idea. Can you try to fix `index.js` in your `node_modules/size-limit`?

If it will work, feel free to send PR.

ai:
Released in 0.19.2

PR #69, opened 2018-08-10
Fix ignore option.

jaydenseric:
Fixes https://github.com/ai/size-limit/issues/68.

I rushed this fix and haven't tested it much.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+# Next
+* Fix [#68](https://github.com/ai/size-limit/issues/68) (by Jayden Seric).
+
 # 0.19.1
 * Use `style-loader` 0.22.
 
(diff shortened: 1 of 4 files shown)


ai:
You need to escape strings in RegExp.

jaydenseric:
What do you mean? I used the same RegExp you already had…

```diff
   if (opts.ignore && opts.ignore.length !== 0) {
     let escaped = opts.ignore.map(i => escapeRegexp(i))
-    let regexp = new RegExp(`^(${ escaped.join('|') })($|/)`)
-    config.plugins.push(new webpack.IgnorePlugin(regexp))
+    config.externals = new RegExp(`^(${ escaped.join('|') })($|/)`)
   }
```

You have push access, so feel free to add polish 😌

I'm pretty desperate for the fix though; the current published version is broken.

ai:
> What do you mean? I used the same RegExp you already had…

What is `escaped` will contain some RegExp control symbols? Like `escaped = 'one.two'`, the result will be `^(one.two)($|/)` which will match to `oneBtwo`, since `.` is any symbol in RegExp.

ai:
> You have push access, so feel free to add polish relieved

OK

ai:
Oops, my mistake. You already escaped string above. :+1:

ai:
Released in 0.19.2

Issue #74, opened 2018-10-27
Add support of dynamic output filename


liunaijia:
For example, the following config will produce wrong size information.

```js
output: {
    path: path.resolve(__dirname, 'dist'),
    filename: '[name].js', // it's a dynamic filename
  },
```

Based on webpack document output filename could be a
- `[hash]`
The hash of the module identifier
- `[chunkhash]`
The hash of the chunk content
- `[name]`
The module name
- `[id]`
The module identifier
- `[query]`
The module query, i.e., the string following ? in the filename
- `[function]`
The function, which can return filename [string]

My `.size-limit.js` config:

```js
module.exports = [
  {
    path: "./src/index.jsx",
    config: "./webpack.config.js",
    limit: "500 KB",
  }
]
```

I know I can set `webpack: false` in size-limit config to calculate size based on the emitted files, but it would be great to build this feature into size-limit, I suppose it also paves the way for supporting multiple entry points.

ai:
Sure. Do you have an idea how to implement it?

liunaijia:
I guess `extractSize` function could read `state.toJson().entrypoints` which has entry points and their assets mapping information, and assets size could be parsed from `state.toJson().assets`.

```
"entrypoints": {
    "main": {
      "assets": [
        "runtime~main.js",
        "runtime~main.js.map",
        "vendors~main.js",
        "vendors~main.js.map",
        "main.js",
        "main.js.map"
      ],
      ...
    },
    "anotherentry": {
      "assets": [
        "runtime~anotherentry.js",
        "runtime~anotherentry.js.map",
        "anotherentry.js",
        "anotherentry.js.map"
      ],
      ...
    }
  },
```

ai:
We can add `entry: 'main'` for this case.

Do you want to make PR to save your name in project history?

ai:
> main is just a default name, it could be changed in webpack config.

Yeap, this is why we will have `entry` field in Size Limit config to allow user to change it according their Webpack config.

```
module.exports = [
  {
    entry: "main",
    config: "./webpack.config.js",
    limit: "500 KB",
  }
]
```

PR #75, opened 2018-10-27
Support webpack multiple entry points and dynamic output filename

liunaijia:
This PR aims at adding support webpack multiple entry points and dynamic output filename which was discussed at https://github.com/ai/size-limit/issues/74.


liunaijia:
When webpack config file is given in size-limit config neither `path` nor `entry` is needless. Because customized webpack `entry` will be used in such case.

So either `path` or `config` is required in size-limit config.
- When `path` is given, size-limit will use it as entry point, it's suitable for the applications without webpack config.
- When `config` is defined, size-limit will use the given webpack config file to bundle and measure the size of output files.

ai:
> When config is defined, size-limit will use the given webpack config file to bundle and measure the size of output files.

What if user will want to check size of one entry from many?

test/index.test.js
```diff
@@ -100,3 +100,60 @@ it('uses custom webpack config', () => {
...
+      expect.assertions(1)
+      return getSize(null, {
+        config: fixture(`webpack-multipe-entry-points/webpack.config`),
+        entry: 'moduleBad'
+      }).then(() =>
+        expect(true).toBe(false)
+      ).catch(error =>
+        expect(error.message).toBe(
+          'Cannot find entry points moduleBad, ' +
+          'available entry points are moduleA,moduleB,moduleC')
```
ai: `,` → `, ` in entries list

ai:
Great. The last fixed:

You need to add a test in `cli.js` that `entry` could be loaded from Size Limit config.

And you can add validation for `entry` field in config https://github.com/ai/size-limit/blob/master/cli.js#L105

test/index.test.js
liunaijia: Done.

liunaijia:
@ai Thanks for your feedback, I've addressed them.

ai:
Awesome. I will merge it today or tomorrow (sorry, need to prepare for conference)

ai:
Released in 0.21. Thanks for work.

PR #82, opened 2018-12-25
fix(css-loader): remove minimize option

RedTn:
Size limit is throwing an error on my project when I upgrade css-loader to ^2.0.1
image

Seems minimize was already removed as an option in css-loader in [v1.0.0](https://github.com/webpack-contrib/css-loader/releases/tag/v1.0.0), but only in v2.0.0 did they add verification of illegal parameters.

Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,12 +23,14 @@
     "ci-job-number": "^0.3.0",
     "compression-webpack-plugin": "^2.0.0",
     "cosmiconfig": "^5.0.6",
-    "css-loader": "^1.0.1",
+    "css-loader": "^2.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -59,6 +61,7 @@
   "scripts": {
     "spellcheck": "yaspeller-ci *.md",
     "lint": "eslint-ci *.js test/{fixtures/**/,}*.js",
+    "lint-fix": "eslint --fix *.js test/{fixtures/**/,}*.js",
```
ai: It is not cool to change build tools and fix the bug.

PR #86, opened 2019-03-28
Add a tip about CSS files' size calculation

### LINKED ISSUE #85: Size of CSS is calculated wrong when `webpack` option is `true`
Hey guys,

I'm not sure that it's a bug, and now I'm here just to ask.

When I create an empty project:

```json
{
  "name": "test",
  "devDependencies": {
    "size-limit": "^0.21.1"
  },
  "size-limit": [
    {
      "path": "file.css",
      "limit": "5 KB"
    }
  ]
}
```

and add there an empty CSS file:

```
$ ls -la file.css
-rw-r--r--  1 user  group  0 Mar 28 13:50 file.css
```

and then run `size-limit` I get this:

```
Package size: 2.27 KB
Size limit:   5 KB
With all dependencies, minified and gzipped
```

As I understand the root of the problem is webpack's config and loaders that size-limit uses for calculating the size. When I disable webpack adding `webpack: false` to package.json, everything is fine:

```
Package size: 0 B
Size limit:   5 KB
With all dependencies, minified and gzipped
```

But I'm not sure that it's a right way to calculate size. I mean, as far as I understand, README says that:

```
Webpack inside Size Limit is very useful for small open source library. But if you want to use Size Limit for application, not open source library, you could already have webpack to make bundle.

In this case you can disable internal webpack.
```

So, technically, I started to create a library, but without webpack. And according to README my option is to use default value (`webpack: true`). Am I wrong?

Anyway, I don't think that was a right output for an empty file.

igoradamenko:
Fixes #85.

Anyway, @ai, do you know, maybe there is a way to subtract size of style-loader or even don't add it to the bundle? It looks like an annoying side effect that shouldn't exist in a such library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,10 +215,9 @@ In this case you can disable internal webpack:
  ],
 ```
 
-Also you should disable `webpack` option when you use Size Limit for tracking 
-size of CSS files. Otherwise you will get wrong numbers, because webpack inserts 
-`style-loader` runtime into the bundle, and this loader adds ≈2 KB 
-to the calculated size.
+If you use Size Limit to track size of CSS files only, you should set `webpack: false`.
+Otherwise you will get wrong numbers, because webpack inserts `style-loader` runtime 
+into the bundle, and this loader adds ≈2 KB to the calculated size.
 
 
 ## JavaScript API
--- a/README.md
+++ b/README.md
@@ -215,6 +215,11 @@ In this case you can disable internal webpack:
  ],
 ```
 
+Also you should disable `webpack` option when you use Size Limit for tracking 
+size of CSS files. Otherwise you will get wrong numbers, because webpack inserts 
+`style-loader` runtime into the bundle, and this loader adds ≈2 KB 
+to the calculated size.
+
 
 ## JavaScript API
 


README.md
```diff
@@ -215,6 +215,11 @@ In this case you can disable internal webpack:
  ],
 ```
 
+Also you should disable `webpack` option when you use Size Limit for tracking
```
ai: 1. `Also,` could be removed.
2. “disable `webpack` option” → “set `webpack: false`”

README.md
```diff
@@ -215,6 +215,11 @@ In this case you can disable internal webpack:
  ],
 ```
 
+Also you should disable `webpack` option when you use Size Limit for tracking 
+size of CSS files. Otherwise you will get wrong numbers, because webpack inserts
```
ai: We should explain that you should disable Webpack only if you are tracking _only_ CSS files

README.md
igoradamenko: Fixed.

README.md
igoradamenko: I rehashed it a bit. Check, please.

(And feel free to squash my commits.)

Issue #94, opened 2019-04-25
Output to file


dotellie:
It would be nice if there were a command parameter one could use to save the output of size-limit to a machine readable file. My primary reason for wanting this would be to utilize [GitLab's metrics report feature](https://docs.gitlab.com/ee/ci/metrics_reports.html) but I'm sure there are many other use cases that can come from doing something like this.

ai:
Good idea. I will ask my friends to implement it.

ai:
What do you think about this output?

```bash
$ npx size-limit --json
[
  {
    name: "index.js",
    passed: true,
    size: 1024,
    loading: 0.05,
    running: 0.1
  }
]
```

dotellie:
Yep, that could totally work! Seems very flexible too 😄

ai:
I created a task (it is our Russian board for open source tasks) and will promote it on Twitter.
https://cultofmartians.com/tasks/size-limit-json.html

Hope, people will fix it on this weekend on next one.

Feel free to send PR if you want it earlier.

ai:
Fix https://github.com/ai/size-limit/pull/97 by @vostrik and released in 1.2.0 version.

PR #97, opened 2019-04-27
feat: add reporter layer to support JSON output

vostrik:
Added  JSON output with `--json` argument.

Also added reporter layer to extend in future base reporters. It can be useful in integrations to such tools like Teamcity, etc.

#94


ai:
I like your architecture. But let’s reduce the number of classes. `JsonReporter` and `HumanReporter` will be enough.

runner.js
```diff
@@ -442,10 +450,10 @@ async function main () {
       }
       fix += chalk.bold(configPath)
     }
-    output += '\n  ' + chalk.yellow(fix)
+    hint = chalk.yellow(fix)
   }
 
-  process.stdout.write('\n' + output + '\n')
+  reporter.logResults({ results, hint })
```
ai: You do not need objects here. Just use arguments `reporter.logResults(results, hint)`

runner.js
```diff
@@ -486,6 +494,8 @@ main().catch(e => {
     msg = e.stack
   }
 
-  process.stderr.write(`${ chalk.bgRed(' ERROR ') } ${ chalk.red(msg) }\n`)
+  reporter.error({
+    message: `${ chalk.bgRed(' ERROR ') } ${ chalk.red(msg) }\n`
```
ai: Should we move ` ERROR ` to a reporter?

vostrik: Yeah, I think so. Also in `warn` method?

reporters/std-json-reporter.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  error ({ message }) {
+    super.error({
+      message: this.format({ message: { error: message } })
+    })
+  }
+}
+
+module.exports = {
```
ai: We can directly export class here

runner.js
ai: Yeap

reporters/std-json-reporter.js
vostrik: Ok. 
Do you think that named export to save naming consistence in this case is bad practice?

ai: I am exporting function directly without object wrap https://github.com/ai/size-limit/blob/master/index.js#L255

So for consistency, we should do the same with new files.

runner.js
vostrik: Done in all methods

runner.js
vostrik: Done for `warn()` and `error()` in `HumanReporter`

reporters/std-json-reporter.js
vostrik: Done for all new modules

ai:
I simplified code even more 27891e6

PR #99, opened 2019-04-30
Fixes height of images in readme

artplan1:
Remove fixed `height` props from images to make its look ok on small screens

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,15 +16,14 @@ if the cost exceeds the limit.
   to know if a pull request adds a massive dependency.
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


ai:
Removing `height` will force a page to jump during image loading.

Can we use `intrinsicsize`?

artplan1:
I see.

I tried `intrinsicsize` and looks like GH trims it, also it trims `style` too so no `min-height` here

ai:
OK. Mobile design is more important.

Issue #101, opened 2019-05-02
Reporting: configurable reporter format


grachpower:
# 🚀Feature request
### Description
Now we can not select any output format.

### Alternatives you've considered
Provide any size-limit configuration with reporter type such as `html`, `plain`, `json`, etc.
```
+ "size-limit": [
+   {
+     ...
+     "path": "dist/app-*.js",
+     "reporter": "json"
+   }
+ ],
```

### Benefits
Customers like me can use any external build loggers to manage build statics

ai:
I think it is better to specify format in CLI call:

```json
"scripts: {
  "test": "jest && size-limit --json"
}
```

It could be useful to have different output formats for different cases:

```js
"scripts: {
  "ci:test": "jest -i && size-limit --json",
  "test": "jest && size-limit"
}
```

But you can change my minds by use cases, where the format in config could be more useful.

grachpower:
CLI json output looks good, thanks! it'l be helpful description in readme

ai:
Yeap. Can you send PR to docs?

PR #102, opened 2019-05-02
Add json format output description

grachpower:
Related to https://github.com/ai/size-limit/issues/101


ai:
Usage steps is not the best place. This section is for very strict instructions and any point should have use case.

I think it is better to add separated section after “JS API”. In this section we can show some useful example where to use JSON output.

PR #108, opened 2019-05-19
Add mobx-state-tree case

pustovalov:
https://github.com/mobxjs/mobx-state-tree/pull/1292

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ and show the real cost of all your internal dependencies.
 ## Who Uses Size Limit
 
 * [MobX](https://github.com/mobxjs/mobx)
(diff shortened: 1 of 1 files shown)


ai:
The rules about the Users section. It should be a very big project or good minification step (like −20% of the size).

I have this rule to keep section clear and focused (a big list is worse than small with few most important items).

pustovalov:
Ok then, how is a very big project determined?

ai:
I think it will be better for you just to reduce `mobx-state-tree` size and use % in this section.

Anyway “X uses Size Limit” is just hype driven development. Size Limit should be useful.

pustovalov:
Ok, thx

PR #115, opened 2019-06-06
Use typescript (maybe)

just-boris:
This is not an actual PR, but follow-up to the twitter thread, how could possibly a migration to typescript could look like.

Open for feedback from the community


ai:
Even out of typing/dynamic topic this PR made the system more complicated.

Right now I cam running files directly without any compilation step. This PR made everything much more complicated with complication step (I know about all side projects of it during PostCSS/Autoprefixer development process).

This is why I am closing process. We should not use build step for modern Node.js projects.

src/cli-body.ts
```diff
@@ -1,13 +1,19 @@
...
+  OutputFileEntry,
+  InputFileEntry
+} from './interfaces'
+import { isAbsolute, relative, join, dirname } from 'path'
+import cosmiconfig from 'cosmiconfig'
+import readPkg from 'read-pkg-up'
+import globby from 'globby'
+import yargs from 'yargs'
+import bytes from 'bytes'
+import chalk from 'chalk'
```
ai: Note, that is invalid module syntax. ES modules always must have file extensions in the end.

(This note is just to spread correct ES modules syntax)

just-boris:
```js
import globby from 'globby';
console.log(typeof globby)
```

```
node --experimental-modules test-module.mjs
function
```

it works! Why is this not valid?

ai:
> it works! Why is this not valid?

It is Node.js non-official improvement, which will not work in browsers.

Node.js tries to push the changes to the spec, but discussion is still in process.

BTW, is spec-based `./node_modules/globby/index.js` work without `--experimental-modules` in Node.js 12 (sorry, I am out of laptop right now).

styfle:
I think you are overlooking the fact that `"module": "commonjs"` in the `tsconfig.json` means that TS will convert the import calls to require calls.

ai:
@just-boris `import globby from 'globby'` should not work by spec because Globby use CommonJS modules, not ES modules

PR #122, opened 2019-06-14
Greedy ignore pattern, fixes #121

### LINKED ISSUE #121: Reported size are not (always) correct
Comparing `0.21` vs `1.3`:

### 0.21 - that's a true size
```text
 dist/es5/index.js
  Package size: 7.91 KB

  dist/es2015/sidecar.js
  Package size: 4.31 KB

  dist/es2015/UI.js
  Package size: 2.45 KB
```

### 1.3.6
```text
  dist/es5/index.js
  Package size: 1.47 KB

  dist/es2015/sidecar.js
  Package size: 807 B

  dist/es2015/UI.js
  Package size: 576 B
```

## Another project
Which is __included__ in the project above, ie `sidecar.js` == `project1car.js + project2car.js`.

### 0.21
```text
 dist/cjs/UI.js
  Package size: 2.5 KB

  dist/es2015/sidecar.js
  Package size: 2.95 KB

  dist/es2015/index.js
  Package size: 4.33 KB
```

### 1.3.6
```text
 dist/cjs/UI.js
  Package size: 1.69 KB

  dist/es2015/sidecar.js
  Package size: 2.63 KB

  dist/es2015/index.js
  Package size: 3.79 KB
```
In this case, the difference is not as big, but still present. Probably that's some webpack optimizations, but size-limit has no fixed webpack version set, so it shall be the same between different `size-limit` versions.

---

According to `size-limit` side-car of the following project is about 3kb, while sidecar of a parent project(which imports it), is just 807b. Meanwhile, 0.21 reported the size correctly.

According to `--why` 2 of 4 external packages do not "present", even in a `dev` mode, that's why size is so small - it calculates nothing.

I have no idea about why this is happening, but have to report the issue.

theKashey:
Addresses the problem raised in #121 - having `react` as `peerDep` would add any `react-*` packages to the ignore list, which is almost all packages :)

## Fix
Restore old RegExp based check, keeping the new `externalize` behavior

## Testing
I've added `redux-thunk`, as long it matches the problem, and configured test to handle it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+let escapeRegexp = require('escape-string-regexp')
 let OptimizeCss = require('optimize-css-assets-webpack-plugin')
 let Compression = require('compression-webpack-plugin')
 let Analyzer = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
@@ -98,8 +99,10 @@ function getConfig (files, opts) {
   }
 
   if (opts.ignore && opts.ignore.length !== 0) {
+    let escaped = opts.ignore.map(i => escapeRegexp(i))
+    let ignorePattern = new RegExp(`^(${ escaped.join('|') })($|/)`)
     config.externals = (context, request, callback) => {
-      if (opts.ignore.some(i => request.startsWith(i))) {
+      if (ignorePattern.test(request)) {
         callback(null, 'root a')
       } else {
         callback()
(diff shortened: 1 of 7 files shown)


ai:
Thanks

ai:
Released in 1.3.7

PR #123, opened 2019-06-14
Control over webpack ModuleConcatenationPlugin, fixes #117

### LINKED ISSUE #117: --why and multiple entries
There two issues with `--why`
- it does not support multiple entries, showing results only for the first one. I would like to be able to specify an entry point to "explain why"
- it does not do the work well.

### Let me elaborate on this
First of all - what is the goal of `--why`? Explain "what's inside, and why it's so big". Does it work? Nooooopppeee, as long as you are size-limiting __esm__(aka `modules:false`) bundle, which got properly tree shaked(👍), and concatenated(👎), with all the details lost.
- [ ] disable concatenation for `--why`

Secondary, as I asked above, I want to _size-limit_ multiple entries, and I would like to call them by names, or even just use these entry points as `path`
```diff
  {
-    path: "dist/es2015/index.js",
+    path: ".", // look for package.json/index.js -> resolve target file -> do the jon
+    path: "stuff", // look for stuff.js, stuff/package.json, stuff/index.js and so on
     limit: "2.0 KB"
  },
```
The "file" behind `path` could be resolved via `Module._resolveFilename`

### Outcome
You will measure not you what want, but what user will consume,

- [ ] WYSIWYG size-limit. Yep, it's really "what you see is what you get"

theKashey:
Without this option `--why` is almost useless for tree-shakable (and thus _concatenatable_) modules - see #117.
In the same time - disabling concatenation greatly affect result bundle size, thus it should be optional.

### Before
Screen Shot 2019-06-13 at 1 58 05 pm

### Afrer
Screen Shot 2019-06-13 at 1 57 41 pm

### How
- Added a new CLI option
- Added a new API option with _reversed_ sense, to keep all existing customers untouched.

### Tests
It will require the usage of ESM. While this could be a good thing to test, I am not ready to do it as my first-time contribution - I've tested manually.

### DIFF
--- a/cli-body.js
+++ b/cli-body.js
@@ -467,6 +467,7 @@ async function run () {
       analyzer: process.env['NODE_ENV'] === 'test' ? 'static' : 'server',
       bundle: config.bundle,
       output: argv.saveBundle,
+      disableModuleConcatenation: true,
       ignore
     }
     let full = files.reduce((all, i) => all.concat(i.full), [])
--- a/index.js
+++ b/index.js
@@ -63,6 +63,9 @@ function getConfig (files, opts) {
       filename: projectName(opts, files),
       path: getBundlePath(opts)
     },
+    optimization: {
+      concatenateModules: !opts.disableModuleConcatenation
+    },
     module: {
       rules: [
         {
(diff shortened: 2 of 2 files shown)


ai:
Maybe we should enable it by default? It will be OK if `--why` will show a little different size (or we can print a warning in the console).

I am afraid that users will miss this option (nobody reads docs).

theKashey:
It would be a better option to disable some optimizations for `--why` out of the box. If you are ok with it - I'll remove CLI changes and set this option for `--why` by default.

ai:
> If you are ok with it - I'll remove CLI changes and set this option for --why by default.

:+1:

theKashey:
Forgot to revert readme...

ai:
Released in 1.3.7

PR #130, opened 2019-08-15
Readme tweaks: difference between big and small

arturi:
- Added explanation of the difference between “big” and “small” presets.
- Minor grammar tweaks here and there.


ai:
Thanks

PR #145, opened 2019-11-04
Allow to specify chrome executable

sHooKDT:
As for now I have no option to specify chrome executable to use in @size-limit/time plugin. It causes problems for me in CI, because I'm unable to get it up and running on Node Alpine image.

It uses estimo, which uses puppeteer-core (not original puppeteer) inside, so it is not affected by PUPPETEER_* env variables. But as I found out, it passes options down to it.

That said I suggest to introduce CHROME_EXECUTABLE_PATH environment variable. I was also thinking to reuse PUPPETEER_EXECUTABLE_PATH env, but maybe it breaks incapsulation a bit, since we know you are using puppeteer inside.
I don't know how to test is locally for now. So I haven't verified it is working, but maybe you can publish some beta version in order to test it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,6 +376,10 @@ Each section in the config can have these options:
 * **ignore**: an array of files and dependencies to exclude from
   the project size calculation.
 
+Plugin for execution time calculation uses chromium inside.
+If you are using it on CI, you may want to specify your custom chromium binary.
+It's available under `CHROME_EXECUTABLE_PATH` environment variable.
+
 If you use Size Limit to track the size of CSS files, make sure to set
 `webpack: false`. Otherwise, you will get wrong numbers, because webpack
 inserts `style-loader` runtime (≈2 KB) into the bundle.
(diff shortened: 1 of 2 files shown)


ai:
Yeap, we will fix this bug, but let’s think a few extra minutes to find the best way.

1. What is your CI?
2. Why Estimo can’t find Chrome there?
3. Should we fix the issue in Estimo?
4. Should we add familiar `PUPPETEER_EXECUTABLE_PATH` support to Estimo?

sHooKDT:
1. I'm using docker inside Jenkins pipeline. So we should be able to reproduce it using pure Dockerfile and node:erbium-alpine image.
2. 🤷‍♂️I had some friction to get puppeteer up and running and finally solved it with separately installed chromium and manual binary specification. Estimo uses some custom "find chrome mechanism". I think I can dive deep into it a bit later to find out why.
3. Maybe we can add CLI argument there also, but basically it has Node API and passes parameters down to puppeteer, so we're able to specify executable there.
4. WIth 3. said, I don't know. Because you can solve this problem using pure estimo.

ai:
Can I ask you to create PR to estimo with `PUPPETEER_EXECUTABLE_PATH`.

I afraid that another users with the same problem will not be able to find this environment variable. And we need to fix the problem for all estimo users, not just for Size Limit. I will help you in that PR.

sHooKDT:
Sure, I will do it tomorrow and call you there.
Will we need any changes to size-limit then?

ai:
> Will we need any changes to size-limit then?

I think no. Or just estimo version bump.

PR #147, opened 2019-11-15
CLI message: use a plural form of the verb "contain" 

molefrog:
Just a small spelling fix:
> All plugins and presets already **contain** it as own dependency

I was actually thinking about a more informative message:
> As of __<version_number>__ having `size-limit` as a dependency is no longer required.
> You can safely remove it from your package.json since all plugins and presets already contain it as a dependency.
> Read more: https://github.com/ai/size-limit#usage

Let me know what you think.

### DIFF
--- a/packages/size-limit/create-help.js
+++ b/packages/size-limit/create-help.js
@@ -97,7 +97,7 @@ module.exports = process => {
     let { rm } = npmCommands(pkg)
     printError(
       chalk.bgYellow.black(' WARN ') + ' You can remove size-limit dependency',
-      '       All plugins and presets already contains it as own dependency',
+      '       All plugins and presets already contain it as own dependency',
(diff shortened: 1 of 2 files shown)


molefrog:
P.S. I tried to update test snapshots by using `jest` command, but it resulted in some other keys being updated as well:
```diff
- \\"webpackOutput\\": \\"/tmp/\\",
+ \\"webpackOutput\\": \\"/var/folders/62/hp0h_c990v5csr0n4s6ndgl80000gn/T/\\",
```

I think it has something to do with the fact that I run it on a Mac.

ai:
I will try to fix test snapshots later

ai:
I hope it will fix tests 419143a

ai:
Released in 2.2.1. Thanks.

Issue #144, opened 2019-10-29
Add brotli size computation


rifler:
about 90% of browsers support brotli encoding
https://caniuse.com/#search=brotli

ai:
1. What use case do you have to compare JS by Brotli?
2. What API do you suggest?

rifler:
1. I don't understand your question
2.
a) we can remove gzip and show only brotli
b) show both sizes:
```
  Size (gzip):           20.52 KB with all dependencies, minified and compressed
  Size (brotli):         18.76 KB with all dependencies, minified and compressed
  Loading time (gzip):   411 ms   on slow 3G
  Loading time (brolti): 395 ms   on slow 3G
  Running time:          80 ms    on Snapdragon 410
  Total time (gzip):     490 ms
  Total time (brotli):   470 ms
```
But it looks little bit noisy. We can play with display format

ai:
Remove gzip and show only Brotli is a good idea

ai:
Do you want to prepare PR? But we will need an option to show gzip numbers as well.

rifler:
I'll try soon

binjospookie:
`createBrotliCompress` has been added in: v11.7.0. I found that size-limit's CI config file (.travis.yml) includes Node.js versions below (8, 10). 
Does it mean deprecation versions below 11.7.0 or just showing a message like "Brotli report will be available after upgrading node above v11.7.0)?

ai:
Warning will be enough (since it is opt-in feature)

ai:
Released in 3.0

PR #151, opened 2020-01-05
feat: add brotli support #144

binjospookie:


packages/file/index.js
```diff
@@ -24,8 +45,22 @@ function gzipSize (path) {
 
 let self = {
   name: '@size-limit/file',
-  async step60 (config, check) {
+  async step60 (config, check, nodeVersion = process.version) {
```
ai: steps API do not support passing `nodeVersion` and it is not a big reason to add this argument to steps API.

It will be better to mock `process.version` in specific test.

packages/file/index.js
```diff
@@ -1,13 +1,34 @@
-let { createGzip } = require('zlib')
+let chalk = require('chalk')
+let { gte } = require('semver')
+let { constants, createBrotliCompress, createGzip } = require('zlib')
 let { promisify } = require('util')
 let fs = require('fs')
 
 let stat = promisify(fs.stat)
+let brotliRequiredNodeVersion = 'v11.7.0'
```
ai: It should be `const BROTLI_NODE_VERSION`

packages/file/index.js
```diff
@@ -24,8 +45,22 @@ function gzipSize (path) {
...
+    if (check.brotli === true) {
+      let isBrotliSupported = gte(nodeVersion, brotliRequiredNodeVersion)
+
+      if (isBrotliSupported) {
+        check.size = await sum(files, async i => brotliSize(i))
+
+        return
+      }
+      process.stderr.write(
+        chalk.yellow(`Brotli required node >= ${ brotliRequiredNodeVersion }\n`)
```
ai: In this case, it will be better to stop with an error, since we can’t show size with Brotli.

packages/file/index.js
binjospookie: fixed

packages/file/index.js
binjospookie: fixed

packages/file/index.js
binjospookie: fixed

packages/file/index.js
```diff
@@ -26,6 +47,20 @@ let self = {
...
     let files = check.bundles || check.path
+
+    if (check.brotli === true) {
+      let isBrotliSupported = gte(process.version, BROTLI_NODE_VERSION)
+
+      if (isBrotliSupported) {
+        check.size = await sum(files, async i => brotliSize(i))
+        return
+      }
+      process.stderr.write(
```
ai: Size Limit has a special system for [error messages](https://github.com/ai/size-limit/blob/master/packages/size-limit/size-limit-error.js).

You need to:
1. Add text to `MESSAGES`
2. Use “how to solve” message instead of “what is the problem”. For instance, “Update your Node.js to version >= X to use Brotli”.
3. `throw` an error from a module (it will keep your from requiring `chalk` and make the test much cleaner).

binjospookie: Sounds good

ai:
Thanks. Released in 3.0 (with Node.js 8 drop).

binjospookie:
@ai thanks.

Issue #154, opened 2020-01-28
Yarn v2 support


pustovalov:
Hey,

I have this declaration for size-limit in `package.json`

```
"devDependencies": {
  "@size-limit/preset-app": "^3.0.0",
  "@size-limit/webpack": "^3.0.0"
}
```

with yarn v1 everything worked well, but with yarn 2 and PnP, I got this an error message:

```
[~/site-new]$ yarn size                                                                                                                                               
Usage Error: Couldn't find a script named "size-limit".

$ yarn run <scriptName> ...
```

```
[~/site-new]$ yarn bin                                                                                                                                               
➤ YN0000: husky-run
➤ YN0000: husky-upgrade
➤ YN0000: lint-staged
➤ YN0000: prettier
➤ YN0000: sass
➤ YN0000: webpack
➤ YN0000: webpack-bundle-analyzer
➤ YN0000: webpack-cli
➤ YN0000: webpack-dev-server
➤ YN0000: Done in 0.04s
```

After I changed the declaration to:
```
"devDependencies": {
  "@size-limit/preset-app": "^3.0.0",
  "@size-limit/webpack": "^3.0.0",
  "size-limit": "^3.0.0"
}
```

```
[~/site-new]$ yarn size                                                                                                                                               
 WARN  You can remove size-limit dependency
       All plugins and presets already contain it as own dependency
       yarn remove size-limit

[~/site-new]$ yarn bin                                                                                                                                                
➤ YN0000: husky-run
➤ YN0000: husky-upgrade
➤ YN0000: lint-staged
➤ YN0000: prettier
➤ YN0000: sass
➤ YN0000: size-limit
➤ YN0000: webpack
➤ YN0000: webpack-bundle-analyzer
➤ YN0000: webpack-cli
➤ YN0000: webpack-dev-server
➤ YN0000: Done in 0.05s
```

pustovalov:
as I understand yarn 2 applies binaries only from first level dependencies, discussion with the maintainer of yarnpkg:
![dis](https://user-images.githubusercontent.com/1568885/73442008-ae630a80-435c-11ea-87bb-49fb8493fedd.jpg)

ai:
We can change Install instruction to explicitly ask to add `size-limit` to dependencies

pustovalov:
I can make a PR for these changes, today or tomorrow

PR #156, opened 2020-01-30
Yarn 2 support

pustovalov:
Related to https://github.com/ai/size-limit/issues/154


ai:
CI fails because `packages/size-limit/create-help.js:89` line was not covered with tests

ai:
Do we need a major release? Or it will just print warning?

pustovalov:
It will print a warning, if user doesn’t have a size-limit in dev dependencies 
https://github.com/ai/size-limit/pull/156/files#diff-ba32842adaed10794b6c66a24aaea6a4R120

pustovalov:
Do we still need this dependency?
https://github.com/ai/size-limit/blob/master/packages/file/package.json#L21

ai:
Released at 3.0.1.

Yeap, seems like we forget to move `size-limit` from `dependencies` to `peerDependencies`. Can you send another PR?

ai:
I will fix `dependencies`

Issue #161, opened 2020-02-12
Add a way to test tree-shaking size


bschlenk:
This may be out of scope for this project, but it would be really cool to be able to test out tree-shaking benefits with this tool. If there was a way to configure a file and an export to pull from that file, it could be a great way to make sure tree shaking is set up properly.

ai:
We can add `import` directive to the config like:

```json
  "size-limit": [
    "import": "import { method } from '.'",
    "limit": "15 KB"
  ]
```

Will, it fit your task?

Do you want to send PR?

bschlenk:
Yeah, that would be really cool! I'll take a look at it this weekend and see if I can get you a PR.

bschlenk:
I started looking into this, but wanted your advice on the best way to implement. In order for this to work, it would have to write a new file to use as the webpack `entry` file. Where would be the best place to write this file? In the temp directory that gets created [here](https://github.com/ai/size-limit/blob/master/packages/webpack/index.js#L47)?

ai:
> Where would be the best place to write this file? In the temp directory that gets created here?

Yeap, it is a good candidate for this temp file

bschlenk:
This is harder than I thought. If I create a file with just `import { method } from '.'`, webpack treeshakes everything out because `method` is never called. The config could be required to use the import as well, like `import { method } from '.'; method()`, but that doesn't feel right. Any thoughts?

ai:
Another idea:

Config:

```js
  "size-limit": [
    {
      "path": ".",
      "import": "a, b",
      "limit": "15 KB"
    }
  ]
```

In `entry` file you put:

```js
import { IMPORT } from "PATH"
console.log(IMPORT)
```

So, in this case, you will have:

```js
import { a, b } from "."
console.log(a, b)
```

bschlenk:
Didn't see there was already a similar issue. Should I close this in favor of https://github.com/ai/size-limit/issues/111 ?

ai:
Yeap, let’s close this one

Issue #111, opened 2019-05-28
Add custom import


ai:
```js
  import: "import { List } from 'lib'"
```

https://twitter.com/brian_d_vaughn/status/1133378518820069376

theKashey:
Was just going to create the same issue - `scenario-based size measurements`.
Like I am going to use the library like `this` - what would be the impact.

theKashey:
Not promising anything, but I need this feature so maybe.
Do you have API in mind?
I could see two different posibilities:

- simple. Just specify imports. Ie `import default, {List} from "dist/es2015/index.js";`
```js
  {
    path: "dist/es2015/index.js",
    imports: ['default', 'List'], // import "this" from "that"
    limit: "4 KB"
  },
```
- complex (but very interesting) - use `npm pack` to pack the package, and use it as a package, so all endpoints and all `imports` would be 100% real.
```js
  {
    scenario: "
      import { List } from 'myLib";
      import { Tree } from 'myLib/endpoint';
    ",
    limit: "4 KB"
  },
```

theKashey:
It would make `size-limit` a bit slower, as long as `npm pack` might trigger `build`

theKashey:
if you want to `import from 'XXX'`, especially `XXX/yyy` - you have to pack and unpack your stuff. However, we might use webpack alias to bound 'XXX' to the current dir.

theKashey:
😅

ai:
@bvaughn @theKashey we have a solution to test the library size with tree-shaking.

Thanks to @bschlenk. We will release it when we will finish docs,

https://github.com/ai/size-limit/pull/166

ai:
The current API:

```js
  "size-limit": [
    {
      "path": ".",
      "import": "a, b",
      "limit": "15 KB"
    }
  ]
```

Which will create a test:

```js
import { a, b } from "."
console.log(a, b)
```

bschlenk:
I just realized that if `import` takes a string then there isn't really a way to specify the default export. From https://github.com/ai/size-limit/issues/111#issuecomment-525556388, if `import` took an array then we can special case `default` to mean the default export. Although, could a module do `export const default = 'someting'`, in which case they'd never be able to test that one?

[...]

PR #166, opened 2020-02-26
Add import option to check config

bschlenk:
This option allows specifying a specific exported object to be imported
from the configured path, allowing testing for tree shaking.

Closes ~~https://github.com/ai/size-limit/issues/161~~ https://github.com/ai/size-limit/issues/111

TODOs:

- [ ] documentation
- [ ] config validation?
- [ ] verify sizes reported by test

### DIFF
--- a/packages/size-limit/get-config.js
+++ b/packages/size-limit/get-config.js
@@ -14,6 +14,7 @@ let OPTIONS = {
   config: 'webpack',
   webpack: 'webpack',
   ignore: 'webpack',
+  import: 'webpack',
   gzip: ['webpack', 'file'],
   running: 'time',
   disableModuleConcatenation: 'webpack',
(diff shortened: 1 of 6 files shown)


ai:
1. Yeap, we need docs. Can I ask you to add docs as an English native speaker?
2. We can miss config validation for now.
3. Sizes in the test look good. But you can replace `AAAAA` to some random symbols, it will make a size difference in test bigger.

bschlenk:
1. Sure, I'll update with some docs when I get the time.
2. I can tackle this too - the only validation I can think of is that if import is used, then path cannot be an array
3. Would that be to reduce the amount of gzip compression?

ai:
> Would that be to reduce the amount of gzip compression?

Yeap, `AAAA` is doing good for gzip compression, so we do not see a big difference in `methodA` and `methodB`.

ai:
I forced feature because I need it for my open-source which uses [`dual-publish`](https://github.com/ai/dual-publish/)

bschlenk:
This wasn’t ready yet. Doesn’t support default import.

ai:
I added docs and default import. It was easy. You did a main work.

PR #167, opened 2020-03-13
implemented watch mode

jayhoney:
I've trying to implement this as simple as I could do it. So it looks like enough for the requirements maybe it's not that good inside, so let's discuss what we can do better :)


packages/size-limit/run.js
```diff
@@ -43,10 +50,20 @@ module.exports = async process => {
 
     config = await getConfig(plugins, process, args, pkg)
 
-    await calc(plugins, config, ora)
+    if (hasArg('--watch')) {
```
ai: We need to call `mainCalc` first before starting to watch.

packages/size-limit/run.js
```diff
@@ -43,10 +50,20 @@ module.exports = async process => {
...
+      let watcher = chokidar.watch([
+        '**/*.js',
+        'package.json'
+      ], {
+        ignored: '**/node_modules/**'
+      })
+      watcher.on('change', async () => {
```
ai: If the user presses `Ctrl+S` twice very fast, chokidar will start 2 Size Limit’s `mainCalc` in parallel.

We need a way to throttle it. We can look at a solution in chokidar or copy my solution from [print-snapshots](https://github.com/ai/print-snapshots/blob/master/watch.js#L40).

ai:
Looks good. Only few small things to fix.

packages/size-limit/run.js
jayhoney: I've fixed this. The only one question. Why do you set delay to 200ms? Do you need really wait so long or it's enough to wait next cycle like ( setTimeout(() => ..., 0) )?

packages/size-limit/run.js
jayhoney: Yeah, you're right! Fixed

packages/size-limit/run.js
ai: You are right. We need to wait for full circle. `print-snapshots` wait for full circle and 200ms.

https://github.com/ai/print-snapshots/blob/master/watch.js#L14-L16

ai:
Great work!

Can you remind me of your Twitter account (it will be hard to find in mentions)?

jayhoney:
> Great work!
> 
> Can you remind me of your Twitter account (it will be hard to find in mentions)?

Oh, thank you! https://twitter.com/your_bulichka

ai:
Thanks. Released at 4.4.

Do you want to add another contribution to your GitHub and replace `nodemon` with your own `size-limit --watch` [in Nano ID](https://github.com/ai/nanoid/blob/master/package.json#L17)?

What name I should use in [task at Cult of Martians](https://cultofmartians.com/tasks/size-limit-watch.html#task)?

jayhoney:
Yes, I'd like to continue implementing things that helps other developers and even big teams :)

I guess my github nickname, "jayhoney" would be nice

ai:
Done https://cultofmartians.com/tasks/size-limit-watch.html#task

I am waiting for PR to Nano ID (you just need to remove `nodemon` and use `size-limit --watch` directly)

Issue #175, opened 2020-04-17
Feature proposal:  extend default webpack configuration


toomuchdesign:
Hi @ai,
I was wondering whether the maintainers of this library have strong opinions against allowing users to **freely extend default webpack configuration** like, to mention one, [storybook does](https://storybook.js.org/docs/configurations/custom-webpack-config/).

It could consist of single call to a user-provided function that mutates or recreates webpack's config.

In my specific case, I wanted to add `babel-loader` to strip out `propTypes` declarations with [babel-plugin-transform-react-remove-prop-types](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types) shipped with the library under test, but I think this extra layer could come in handy in a lot of corner cases.

Glad to make a PR if interested.

Thank for your efforts, here. Cheers!
Andrea,

ai:
Why `config` option is not a solution for you https://github.com/ai/size-limit#config

toomuchdesign:
I'd like to fully rely on the default provided by the library in [`webpack/get-config.js`](https://github.com/ai/size-limit/blob/master/packages/webpack/get-config.js) and add something on top of it.

If I don't get it wrong, it might also back us up in case defaults settings get updated in future. 🤞

ai:
Another way. Write a Size Limit plugin, which will runs after `@size-limit/webpack` and change its config. I can give you access to `@size-limit/` npm’s namespace or we can add custom plugins API to Size Limit.

toomuchdesign:
Good point. A custom `size-limit` plugin should provide the necessary grip surface to expand the bundling configuration. Thank you so much!

Closing for now, I'll ping you back if I end up with something worth publishing.
Keep up the great work.

PR #177, opened 2020-04-25
Add webpack-extender plugin

toomuchdesign:
This is a follow up to #175 about extending the default webpack configuration provided by `size-limit` with a custom function call like:

```js
(defaultConfig) => {
  // Return extended Webpack config
}
```
I explored the direction suggested by @ai, and roughly added a **new size-limit plugin** meant to **run after webpack plugin** to extend its configuration.

It's probably a bit overkill, but it basically works. An open point remains: how to provide the path to the configuration extender file (the config solution implemented in this PR is just a quick workaround):

1. Adding a **new `size-limit` option** seems contradicting the philosophy of extensions by adding self-contained plugins
2. Reusing the **existing `config` option** could also work if we added a runtime check to tell whether the provided file exports an object or a function. Which is a bit "meh!".

If option 2 were considered acceptable, this entire `webpack-extender` plugin could be implemented with addition of 2 lines of code [here](https://github.com/ai/size-limit/blob/062c459612590fc0afb98ea5ad523a556e0af684/packages/webpack/index.js#L53).

I'll let to the maintainers the considerations about both implementation and need for such a feature.

I had the opportunity to take a quick look at [some implementation details ](https://github.com/ai/size-limit/blob/master/packages/size-limit/calc.js#L4) of this library and I wanted to express my praise. Clever and funny thing.

Have a bright day


package.json
```diff
@@ -14,6 +14,7 @@
     "packages/dual-publish",
     "packages/size-limit",
     "packages/webpack",
+    "packages/webpack-extender",
```
ai, line 17: What do you think about `webpack-config` name to specify that we are extending Webpack config?

packages/webpack-extender/LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
+
+Copyright 2017 Andrey Sitnik <andrey@sitnik.ru>
```
ai, line 3: You need to put `2020` and your name :)

toomuchdesign:
Hi @ai!
Let's go for the plugin solution, then.

There' still a point I'd like to hear your opinion about: how to provide this plugin with an option to supply the path to the config extender file. Possible options:

1. Add a new global `configExtender` option
2. Tweak the existing `config` option to serve both `webpack` and `webpack-config` plugins
3. Expose a factory function and provide options programmatically (`makeWebpackConfig(options)`)
4. Same as above but exposing a class to stay more idiomatic (`new WebpackConfig(options)`)
5. Anything else?

ai:
On the one hand it will be nice to use existed config.

But in another hand, Size Limit uses a lot of tricky dynamic field, and it may be better to use API like `extender: (config) => config`.

But we in Size Limit config we can put only link to the file. Webpack config must be a JS files, but the most popular Size Limit config place is `package.json`.

toomuchdesign:
Current status:
now we have a new `webpack-config` plugin consuming a `configExtender` global option supposed to point to a JS file exposing a function with signature  `(config) => config`;

Since we're adding a global option anyway, I'm still quite inclined to think that adding a new plugin is currently a bit overkill, but you know better then anyone the direction this library is supposed to take.

Let me know whether there's something you'd like to review (except for docs which I'll add at the end)

## Extra: provide option objects JS API

From what I can see, there's in currently no support for options configurations via JS API.

Since tests mainly rely on those API's, there's a lot of [mocks aware of implementation details](https://github.com/ai/size-limit/blob/062c459612590fc0afb98ea5ad523a556e0af684/packages/webpack/test/index.test.js#L18) in place necessary to run the business logic with different option set.

Could a **3rd option parameter** improve the viable testing strategies (beside allowing a full configuration from JS)?

```js
sizeLimit([plugins], [filePath], {options})
```

ai:
> From what I can see, there's in currently no support for options configurations via JS API.

Nope. We have a JS API which accepts `[files]` or `internalConfig` https://github.com/ai/size-limit/blob/master/packages/size-limit/index.js#L13-L17

> Since we're adding a global option anyway, I'm still quite inclined to think that adding a new plugin is currently a bit overkill, but you know better then anyone the direction this library is supposed to take.

Maybe you are right and adding small option will not be bad for `size-limit/webpack`.

PR #181, opened 2020-05-28
Add newlines between migration guide variants

saitonakamura:
It's easier to read this way, and more clear to what case which command corresponds


ai:
@saitonakamura you also need to update tests by `npx jest -u`

saitonakamura:
Wow, that was quick 😄 K, will do

Issue #184, opened 2020-06-29
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 2 SIGWINCH listeners added to [process].


timocov:
I'm trying to integrate `size-limit` to https://github.com/tradingview/lightweight-charts and run the following script:

```javascript
const sizeLimit = require('size-limit');
const filePlugin = require('@size-limit/file');

// here could be any file instead of dist/lightweight-charts.esm.production.js, e.g. package.json
sizeLimit([filePlugin], ['dist/lightweight-charts.esm.production.js']).then(result => {
	console.log(result);
});
```

I use NodeJS 14.4 if it's related.

If I run with `--trace-warnings`, it outputs the following:

```
(node:22356) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 2 SIGWINCH listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:390:17)
    at process.addListener (events.js:406:10)
    at process.getStderr [as stderr] (internal/bootstrap/switches/is_main_thread.js:135:13)
    at ReadStream.Readable.pipe (_stream_readable.js:656:32)
    at C:\projects\lightweight-charts\node_modules\@size-limit\file\index.js:37:42
    at new Promise (<anonymous>)
    at gzipSize (C:\projects\lightweight-charts\node_modules\@size-limit\file\index.js:35:10)
    at C:\projects\lightweight-charts\node_modules\@size-limit\file\index.js:61:48
    at Array.map (<anonymous>)
    at sum (C:\projects\lightweight-charts\node_modules\@size-limit\file\index.js:11:35)
```

Similar to #146.

Both `size-limit` and `@size-limit/file` are `4.5.2` version.

timocov:
It's interesting that running `size-limit` doesn't cause this warning (with enabled `file` check) 🤔

timocov:
@ai from https://github.com/ai/size-limit/issues/146#issuecomment-651444680:

> Start with Size Limit CLI. Maybe it is a problem of JS APII(which is not really user-friendly)

We use both CLI and JS API, because we need to be able to check that size isn't out of our limits and need to see what's the difference since merge-base. I saw GitHub Action for that, but we can't use GitHub Actions for some reasons, unfortunately - that's why we write in-house script to handle this.

> If you still will have a problem, open another issue

Yep, here we are :)

ai:
Am I right that you have this issue only in JS API?

[...]

PR #186, opened 2020-06-30
Fixed note while using file plugin with enabled gzip

### LINKED ISSUE #185: Using size-limit with just `file` plugin doesn't output `gzipped` as note
It seems that there is no `gzip` plugin referenced in https://github.com/ai/size-limit/blob/3571e79736a5cd7bf9cf245ff7d5ea2e3ac5f9ef/packages/size-limit/create-reporter.js#L86-L87, isn't it? Just guess that in this check we need to have something like `plugins.has('file') && check.gzip !== false`?

timocov:
Fixes #185

### DIFF
--- a/packages/size-limit/create-reporter.js
+++ b/packages/size-limit/create-reporter.js
@@ -83,7 +83,7 @@ function createHumanReporter (process) {
           sizeNote = 'with all dependencies and minified'
         } else if (plugins.has('webpack')) {
           sizeNote = 'with all dependencies, minified and gzipped'
-        } else if (plugins.has('gzip')) {
+        } else if (plugins.has('file') && check.gzip !== false) {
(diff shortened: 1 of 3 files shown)


ai:
Can you update snapshots by calling `npx jest -u`?

timocov:
Yeah, I'm in the process... It was self-confidently to trying to fix it in without checking out the code by modifying a file right on GitHub 😂

timocov:
I'm not sure how the failed tests are related to this PR. Can you help me with that please?

ai:
I re-run the tests and they passed. Seems like it was a random issue.

ai:
I will think we will release it together with a fix for https://github.com/ai/size-limit/issues/184

PR #188, opened 2020-07-11
add check for missing DIR parameter of save-bundle argument

MikeDevice:
I made this PR, because of two reasons. The first one is the command

```
size-limit --save-bundle
```

does't save bundle and trigger any errors. And the second one is

```
size-limit --save-bundle --clean-dir
```

saves bundle to `--clean-dir` folder.


ai:
Thanks :)

ai:
Released in 4.5.4

PR #199, opened 2020-09-04
Add `--hide-passed` flag to only output failed checks

koddsson:
The output can be very long if you have a lot of bundles that you're checking. The `--hide-passed` option will only output check failures.


ai:
I like the idea!

But the argument name is too general and not self-explained. `--hide-passed`?

koddsson:
> But the argument name is too general and not self-explained. `--hide-passed`?

Sounds good 👍🏻

PR #201, opened 2020-09-16
Ignore unknown command line arguments

### LINKED ISSUE #200: Option to filter reported sections or allow passing custom options via command line
I have a monorepo with dozens of libraries. The value exported by the `.size-limit.js` module is built dynamically as an array of sections (i.e. `{name: string, path: string}[]`). Running the script with the `--why` option opens dozens of tabs on my browser. Ideally, I can pick one or more sections for introspection.

I tried passing custom options to modify the generated configuration, but `size-limit` bails when unknown options are passed via the command line. Ignoring unknown options is also a good enough solution for me as that would allow me to use a custom option to filter the results.

ctumolosus:
Fixes #200


ai:
Closing since environment variables is better for hacks

PR #203, opened 2020-10-27
Add --highlight-less flag // closes feature request #202

### LINKED ISSUE #202: Optionally add reporting about lesser size
Now if size is exceeding limit, there is error message:
```
  build/index.js
  Package size limit has exceeded by 9 B
  Size limit: 360 B
  Size:       369 B with all dependencies, minified and gzipped
```

But there is no "hurray" message about lesser size, for example:
```
  build/index.js
  Package size is 5 B less than limit 🎉
  Size limit: 360 B
  Size:       355 B with all dependencies, minified and gzipped
```
(of course this should not cause error)

When I have many files to check against size-limit, and I'm doing some work to decrease library size, trying different approaches over and over, it is sometimes hard to see few bytes difference in `size-limit` report.

Maybe add command line option, or environment variable for that?

victordidenko:
Looks like this in my color scheme:
Screenshot 2020-10-27 at 17 19 23

Closes #202

### DIFF
--- a/packages/size-limit/create-help.js
+++ b/packages/size-limit/create-help.js
@@ -28,13 +28,14 @@ module.exports = process => {
       'Check the real performance cost of your front-end project to users',
       '',
       b('Core options:'),
-      `  ${y('--limit LIMIT')}  Set size or running time limit for files`,
-      `  ${y('--json')}         Show results in JSON format`,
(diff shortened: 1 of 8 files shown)


ai:
Thanks. Released in 4.7.

PR #204, opened 2020-11-10
Emit Webpack stats JSON file

kinday:
This PR utilises `BundleAnalyzerPlugin` to emit Webpack stats file alongside with the bundle when `--save-bundle` is used. It comes handy when one wants to inspect the bundle as [Webpack’s analyser](http://webpack.github.io/analyse) could be used to find out more about reasons why some dependency was included.

### DIFF
--- a/packages/size-limit/test/create-reporter.test.js
+++ b/packages/size-limit/test/create-reporter.test.js
@@ -1,6 +1,3 @@
-let { join } = require('path')
-let { tmpdir } = require('os')
-
 let createReporter = require('../create-reporter')
 
 function results (types, config, isJSON = false) {
@@ -53,9 +50,7 @@ it('renders results', () => {
           time: 3,
           passed: true
         }
-      ],
-      cleanDir: true,
-      saveBundle: join(tmpdir(), `size-limit-test`)
+      ]
     })
   ).toMatchSnapshot()
 })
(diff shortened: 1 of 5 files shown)


ai:
Cool idea!

Can we show something in the terminal output about using this stat file. For instance:

```

Stats files saved to X.
You can review it by https://webpack.github.io/analyse
```

kinday:
Thanks for the nice suggestion.

I’m not sure about the tests though. Should I have made a separate one? Should I have skipped testing for this particular feature altogether?

ai:
> I’m not sure about the tests though. Should I have made a separate one?

Just check that file was created

kinday:
I might not have time for it this week, so don’t consider it abandoned. I'll get back to it as soon as I am able to.

kinday:
Statements coverage does not meet requirement in that case. Are you fine if I mark it as ignored?

packages/size-limit/test/create-reporter.test.js
```diff
@@ -50,7 +53,9 @@ it('renders results', () => {
           time: 3,
           passed: true
         }
-      ]
+      ],
+      cleanDir: true,
+      saveBundle: join(tmpdir(), `size-limit-test`)
```
kinday, line 58: I’ve just realised `createReporter` doesn’t do anything to filesystem, so I don’t need to make this test so complex.

ai:
Thanks.  Released in 4.8.

Issue #205, opened 2020-11-18
Update to webpack v5


jaydenseric:
We should be optimizing our packages for the most recent version of webpack, which is now v5.

Importantly, it understands the package `exports` field:

- https://nodejs.org/api/packages.html#packages_exports
- https://webpack.js.org/guides/package-exports

Also, [it doesn't automatically polyfill the Node.js `process` global for the browser](https://github.com/webpack/changelog-v5#automatic-nodejs-polyfills-removed). Right now, I'm trying to `size-limit` test a package containing React component code like this:

```js
// Allow the display name to be eliminated from production bundles.
if (typeof process === 'object' && process.env.NODE_ENV !== 'production')
  Foo.displayName = 'Foo';
```

Webpack v4 sees the `process` related code, and inserts a massive `./node_modules/process/browser.js` shim that can not be ignored by the `size-limit` config `ignore` option, because that shim is added by webpack and is not imported anywhere in the code I'm testing.

ai:
Yeap. I like to migrate to webpack 5, but do not have time to do this, since spending all my time with Logux State.

Can you send PR?

jaydenseric:
[Logux State](https://github.com/logux/state) 👀

Does only the `@size-limit/webpack` need work? It seems so running `npm ls webpack`:

```
├─┬ @size-limit/preset-small-lib@4.7.0
│ └─┬ @size-limit/webpack@4.7.0
│   ├─┬ css-loader@5.0.1
│   │ └── webpack@4.44.2 deduped
│   ├─┬ file-loader@6.2.0
│   │ └── webpack@4.44.2 deduped
│   ├─┬ optimize-css-assets-webpack-plugin@5.0.4
│   │ └── webpack@4.44.2 deduped
│   ├─┬ style-loader@2.0.0
│   │ └── webpack@4.44.2 deduped
│   └── webpack@4.44.2
```

ai:
Yeap, only `@size-limit/webpack`

PR #206, opened 2020-11-18
Update webpack to v5

jaydenseric:
This PR updates webpack to v5 and fixes https://github.com/ai/size-limit/issues/205 .

A few outstanding issues:

- Why are some of the sizes in the tests negative now?
- The `webpack-bundle-analyzer` dependency that powers the `--why` flag doesn't support webpack v5 yet, see https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/327#issuecomment-729536298 .

Also, I'm not sure we're setting the `mode` that the webpack v5 migration guide says is required now: https://webpack.js.org/migrate/5/#make-sure-to-use-mode .

### DIFF
--- a/packages/webpack/get-config.js
+++ b/packages/webpack/get-config.js
@@ -1,8 +1,7 @@
 let { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer')
-let PnpWebpackPlugin = require('pnp-webpack-plugin')
 let { promisify } = require('util')
 let escapeRegexp = require('escape-string-regexp')
-let OptimizeCss = require('optimize-css-assets-webpack-plugin')
+let CssMinimizerPlugin = require('css-minimizer-webpack-plugin')
 let { join } = require('path')
 let mkdirp = require('mkdirp')
 let fs = require('fs')
@@ -35,13 +34,9 @@ module.exports = async function getConfig (limitConfig, check, output) {
       path: output
     },
     optimization: {
-      concatenateModules: !check.disableModuleConcatenation
-    },
-    resolve: {
-      plugins: [PnpWebpackPlugin]
-    },
-    resolveLoader: {
-      plugins: [PnpWebpackPlugin.moduleLoader(module)]
+      concatenateModules: !check.disableModuleConcatenation,
+      minimize: true,
+      minimizer: ['...', new CssMinimizerPlugin()]
     },
     module: {
       rules: [
@@ -67,14 +62,13 @@ module.exports = async function getConfig (limitConfig, check, output) {
           ]
         }
       ]
-    },
-    plugins: [new OptimizeCss()]
+    }
   }
 
   if (check.ignore && check.ignore.length > 0) {
     let escaped = check.ignore.map(i => escapeRegexp(i))
     let ignorePattern = new RegExp(`^(${escaped.join('|')})($|/)`)
-    config.externals = (context, request, callback) => {
+    config.externals = ({ request }, callback) => {
       if (ignorePattern.test(request)) {
         callback(null, 'root a')
       } else {
@@ -84,6 +78,8 @@ module.exports = async function getConfig (limitConfig, check, output) {
   }
 
   if (limitConfig.why) {
+    if (!config.plugins) config.plugins = []
+
     config.plugins.push(
       new BundleAnalyzerPlugin({
         openAnalyzer: process.env.NODE_ENV !== 'test',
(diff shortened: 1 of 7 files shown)


packages/webpack/package.json
```diff
@@ -21,16 +21,15 @@
   },
   "dependencies": {
     "css-loader": "^5.0.1",
+    "css-minimizer-webpack-plugin": "^1.1.5",
     "escape-string-regexp": "^4.0.0",
     "file-loader": "^6.2.0",
     "mkdirp": "^1.0.4",
     "nanoid": "^3.1.16",
-    "optimize-css-assets-webpack-plugin": "^5.0.4",
```
jaydenseric, line 28: > :warning: For webpack v5 or above please use [css-minimizer-webpack-plugin](https://github.com/webpack-contrib/css-minimizer-webpack-plugin) instead.
> — https://github.com/NMFR/optimize-css-assets-webpack-plugin/commit/41d1e23491328f274d099eff35cbdf53c3dfac8f

packages/webpack/package.json
```diff
@@ -21,16 +21,15 @@
...
-    "pnp-webpack-plugin": "^1.6.4",
```
jaydenseric, line 29: > If you are using Yarn's PnP and the `pnp-webpack-plugin`, we have good news: it is supported by default now. You have to remove it from the configuration.
> — https://webpack.js.org/migrate/5/#clean-up-configuration

packages/webpack/index.js
```diff
@@ -31,11 +31,11 @@ function getFiles (stats, check) {
...
   return Object.keys(entries)
     .reduce((assets, i) => assets.concat(entries[i].assets), [])
-    .map(i => {
+    .map(({ name }) => {
       if (check.webpackConfig.output && check.webpackConfig.output.path) {
-        return join(check.webpackConfig.output.path, i)
+        return join(check.webpackConfig.output.path, name)
       } else {
-        return join(process.cwd(), 'dist', i)
+        return join(process.cwd(), 'dist', name)
```
jaydenseric, line 38: This was causing errors that the paths should be strings, but were objects. How did this ever work? I assume `name` was the original intention?

Also, I tried to refactor this into a cute ternary arrow function return, but it caused multiple ESLint errors as the rules that conflict with Prettier have not been disabled.

packages/webpack/get-config.js
```diff
@@ -67,14 +62,13 @@ module.exports = async function getConfig (limitConfig, check, output) {
...
-    },
-    plugins: [new OptimizeCss()]
+    }
   }
 
   if (check.ignore && check.ignore.length > 0) {
     let escaped = check.ignore.map(i => escapeRegexp(i))
     let ignorePattern = new RegExp(`^(${escaped.join('|')})($|/)`)
-    config.externals = (context, request, callback) => {
+    config.externals = ({ request }, callback) => {
```
jaydenseric, line 71: This solves a webpack v5 warning that the signature has changed.

packages/webpack/get-config.js
```diff
@@ -35,13 +34,9 @@ module.exports = async function getConfig (limitConfig, check, output) {
...
     optimization: {
-      concatenateModules: !check.disableModuleConcatenation
-    },
-    resolve: {
-      plugins: [PnpWebpackPlugin]
-    },
-    resolveLoader: {
-      plugins: [PnpWebpackPlugin.moduleLoader(module)]
+      concatenateModules: !check.disableModuleConcatenation,
+      minimize: true,
```
jaydenseric, line 38: I'm not sure if this should be forced to `true` like this; it depends if we only want it to automatically happen when the `mode` is `production` and not `development`.

packages/webpack/get-config.js
```diff
@@ -35,13 +34,9 @@ module.exports = async function getConfig (limitConfig, check, output) {
...
+      minimizer: ['...', new CssMinimizerPlugin()]
```
jaydenseric, line 39: I don't fully understand the `'...'`, but [the webpack docs say it's the way to do things for webpack v5](https://webpack.js.org/plugins/css-minimizer-webpack-plugin/#getting-started). I think it means allow the default ones, as well as the custom one being provided.

ai:
> Why are some of the sizes in the tests negative now?

You need to update [the empty project size constants](https://github.com/ai/size-limit/blob/master/packages/webpack/index.js#L13-L16) too

> The webpack-bundle-analyzer dependency that powers the --why flag doesn't support webpack v5 yet

It is a big blocker since `--why` is a big feature

packages/webpack/get-config.js
ai: You can remove it and see the tests changes

PR #211, opened 2021-01-30
Look up for plugins in parent package.json

### LINKED ISSUE #153: Respect node require modules resolve mechanism
When I updated size-limit and installed preset on root package.json, I still get the error.
I think because size-limit searches preset dependency using `process.cwd()` and it's not how `require` resolves modules in node. 
image

I can't install preset to everey separate package in my lerna monorepo :(

exah:
Fixes #153 

I do not use `require.resolve`, cause too much logic inside relied on `process.cwd`, instead I decided to find plugins in parent `package.json` dependencies.

### DIFF
--- a/packages/size-limit/load-plugins.js
+++ b/packages/size-limit/load-plugins.js
@@ -14,7 +14,7 @@ class Plugins {
 }
 
 module.exports = function loadPlugins (pkg) {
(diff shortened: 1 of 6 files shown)


ai:
Thanks. Released in 4.9.2.

PR #216, opened 2021-03-07
fail on missed entries

### LINKED ISSUE #213: [Proposal] fall if entry not found
Hi!

I test `./<some-path>/someFileName.js` file with size-limit.

Path to file or file's name can be changed. If I don't change `.size-limit.json` I got a warning `File not found: ./<some-path>/someFileName.js`.

Maybe should add a flag like `--no-warnings` and exit size-limit with error code in this cases?

I can implement it if you're agree.

binjospookie:
Close #213

### DIFF
--- a/packages/size-limit/create-reporter.js
+++ b/packages/size-limit/create-reporter.js
@@ -36,6 +36,19 @@ function createJsonReporter (process) {
   }
 }
 
+function getFixText (prefix, config) {
+  if (config.configPath) {
+    if (config.configPath.endsWith('package.json')) {
+      prefix += ` in ${bold('"size-limit"')} section of `
+    } else {
+      prefix += ' at '
+    }
+    prefix += bold(config.configPath)
+  }
+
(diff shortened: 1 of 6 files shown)


packages/size-limit/create-reporter.js
```diff
@@ -167,7 +168,8 @@ function createHumanReporter (process, isSilentMode = false) {
       }
 
       if (config.failed) {
-        let fix = 'Try to reduce size or increase limit'
+        let fix =
+          'Try to reduce size, increase limit or specify correct file path'
```
ai, line 172: The missed path is a rare case. We should show `specify corrent file path` only if we had this type of error (we can add `config.missed`)

binjospookie: You're right

ai:
Released in 4.10.1

binjospookie:
Thank you

PR #219, opened 2021-03-29
Fix size units

### LINKED ISSUE #218: Size reported incorrectly due to usage of factor 1024
**What happens**  
When the size is `4632 bytes`, size-limit reports `4.52 KB`. However, this size is in `KiB`. This is due to the [bytes package not currently following](https://github.com/visionmedia/bytes.js/issues/26), and [not willing to follow](https://github.com/visionmedia/bytes.js/pull/36), [the standard set forth by ISO, BIPM, and IEC](https://en.wikipedia.org/wiki/Binary_prefix#Adoption_by_IEC,_NIST_and_ISO).

**What should happen**  
size-limit should follow the standard and report `4.63 KB`.

**Solution**  
Use a package like [pretty-bytes](https://www.npmjs.com/package/pretty-bytes) or [bytes-iec](https://www.npmjs.com/package/bytes-iec) which follows the standards and converts sizes properly.

MatthiasKunnen:
I've replaced the [`bytes`](https://www.npmjs.com/package/bytes) packages with [`bytes-iec`](https://www.npmjs.com/package/bytes-iec) which uses the correct units.

I also went through the project and replaced usages of `KB` with `kB` where appropriate.

This fixes #218.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ interactive elements, using React/Vue/Svelte lib or vanilla JS.
     ```sh
     $ npm run size
 
-      Package size: 30.08 KB with all dependencies, minified and gzipped
+      Package size: 30.08 kB with all dependencies, minified and gzipped
     ```
 
 4. Now, let’s set the limit. Add 25% to the current total time and use that as
@@ -134,7 +134,7 @@ interactive elements, using React/Vue/Svelte lib or vanilla JS.
     ```diff
       "size-limit": [
         {
-    +     "limit": "35 KB",
+    +     "limit": "35 kB",
           "path": "dist/app-*.js"
         }
       ],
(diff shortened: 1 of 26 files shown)


ai:
Is it backward compatible? I am OK with limit changes because of the different unit, but will it work with `1 KB` format (not `1 kB`)?

MatthiasKunnen:
Yes, it is. Verify on this runkit: <https://runkit.com/5f46c24ab29d17001ad6a0c2/60620658b40dfa001aa9e18b>.

ai:
WHat do you think of moving to `KiB` in docs?

MatthiasKunnen:
I would not do that. The base 10 notation is much easier to work with, i.e. convert between B <-> kB <-> MB. Linux and MacOS also display file size in base 10. Chrome network monitor also uses the base 10 notation.

ai:
Released in 4.10.2

Issue #221, opened 2021-04-23
Better error message if file is missing


ivanhofer:
It took me a while to realize what was wrong: the file I tried to test was missing, because I forgot to build my library before I ran `npm size-limit`.

I got following error message:
```
✖ Adding to empty webpack project
 ERROR  WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
```

It would be a nice UX improvements when the error could be catched and a "human-readable" error message would be displayed instead. Something like: "Could not test `{{name}}` because file '{{path}}' is missing"

ai:
Please send PR

ivanhofer:
Hi @ai,

i could not get the tests running on my machine for the current main branch, so I can't change and push anything.

My PR would look like:

file: `packages/webpack/get-config.js`
```diff

let writeFile = promisify(fs.writeFile)
+let readFile = promisify(fs.readFile)

const STATIC = /\.(eot|woff2?|ttf|otf|svg|png|jpe?g|gif|webp|mp4|mp3|ogg|pdf|html|ico|md)$/

+async function checkIfFilesExist(limitConfig, check) {
+  const filesToCheck = []
+  for (const file of check.files) {
+    filesToCheck.push(file)
+  }
+  if (check.path) {
+    filesToCheck.push(join(limitConfig.cwd, check.path))
+  }
+
+  for (const file of filesToCheck) {
+    await readFile(file)
+      .catch(() => { throw Error(`Could not test '${check.name}' because file '${file}' is missing`) })
+  }
+}
+
module.exports = async function getConfig(limitConfig, check, output) {
+  await checkIfFilesExist(limitConfig, check)
+
  if (check.import) {
```

ai:
1. You need to use `fs.existsSync` rather than `readFile`. `existsSync` is faster.
2. Send PR and run tests on CI
3. This simple fix does not cover all cases. It falls on webpack’s `supports ignore` test.

ivanhofer:
I now have disabled all git-hooks locally to be able to push to this repository.
I will wait until the CI workflow is approved

ai:
> I now have disabled all git-hooks locally to be able to push to this repository.

What error did you have?

ivanhofer:
I get o lot of different errors:

` Error: Failed to launch the browser process!
[...]

PR #222, opened 2021-05-04
display error message if file is missing

ivanhofer:
fixe: Better error message if file is missing #221


ai:
I am closing since test are still falling

PR #226, opened 2021-05-21
add `modifyWebpackConfig` config option

phryneas:
So, we want to use this for `@reduxjs/toolkit` which will have multiple entry points from the next version that reference each other, so we need to add the `NormalModuleReplacementPlugin`.

Rather than creating a bunch of fixed webpack configs that don't work with `--why`, this allows to just do some last-minute changes to the webpack config.

Hope you'll accept this - it would be a great help for us :slightly_smiling_face: 

### Example usage:

Usage in our case looks like this:
```js
function withRtkPath(path) {
  return (config) => {
    const rtkPath = path.replace(
      /(query\/rtk-query|query\/react\/rtk-query-react)/,
      'redux-toolkit'
    )
    config.plugins.push(
      new webpack.NormalModuleReplacementPlugin(
        /@reduxjs\/toolkit/,
        join(__dirname, rtkPath)
      )
    )
    return config
  }
}
// ...
    {
      name: `createApi (${path})`,
      path,
      import: '{ createApi }',
      modifyWebpackConfig: withRtkPath(path),
      ignore: ['react'],
    },
```
with `path` being from the array
```js
const queryEntryPoints = [
  'dist/query/rtk-query.cjs.production.min.js',
  'dist/query/rtk-query.esm.js',
  'dist/query/rtk-query.modern.production.min.js',
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,6 +505,7 @@ Each section in the config can have these options:
 * **config**: a path to a custom webpack config.
 * **ignore**: an array of files and dependencies to exclude from
   the project size calculation.
+* **modifyWebpackConfig**: (.size-limit.js only) function that can be used to do last-minute changes to the webpack config, like adding a plugin
 
 If you use Size Limit to track the size of CSS files, make sure to set
 `webpack: false`. Otherwise, you will get wrong numbers, because webpack
--- a/README.md
+++ b/README.md
@@ -505,7 +505,8 @@ Each section in the config can have these options:
 * **config**: a path to a custom webpack config.
 * **ignore**: an array of files and dependencies to exclude from
   the project size calculation.
-* **modifyWebpackConfig**: (.size-limit.js only) function that can be used to do last-minute changes to the webpack config, like adding a plugin
+* **modifyWebpackConfig**: (.size-limit.js only) function that can be used to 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -505,6 +505,7 @@ Each section in the config can have these options:
 * **config**: a path to a custom webpack config.
 * **ignore**: an array of files and dependencies to exclude from
   the project size calculation.
+* **modifyWebpackConfig**: (.size-limit.js only) function that can be used to do last-minute changes to the webpack config, like adding a plugin
```
ai, line 508: Add `.` to the end and keep 80 symbols per line limit

phryneas: better? :)

ai: Perfect =^_^=

ai:
We need to fix test coverage

phryneas:
Yeah, I didn't realize you had that strict test coverage in there. 
I can try to add a test tomorrow. 
Going to bed right now, it's late over here ^^

phryneas:
Unfortunately, the tests don't run for me locally, but I think this should do it.

ai:
Thanks. Released in 4.11.

Issue #232, opened 2021-06-09
Reduce dependencies


ai:
https://npm.anvaka.com/#/view/2d/size-limit

Windrushfarer:
Hi! Do you have a vision of the expected result? How many dependencies you think would be enough?

ai:
I think we could replace `read-pk-up` and `ora` huge sub-dependencies branches by replacing them with small built-in functions or with a few zero-dependencies libraries.

PR #234, opened 2021-06-13
Replace read-pkg-up lib by the custom js file

4en1x:
read-pkg-up replacement by custom js file. #232


packages/size-limit/read-pkg-up.js
```diff
@@ -0,0 +1,42 @@
+const path = require('path')
+const { promisify } = require('util')
+const fs = require('fs')
+
+const readFileAsync = promisify(fs.readFile)
```
ai, line 5: Use `fs/promises` instead

packages/size-limit/read-pkg-up.js
```diff
@@ -0,0 +1,42 @@
...
+
+const readPkg = async cwd => {
```
ai, line 7: Use `let` for code style consistency

packages/size-limit/read-pkg-up.js
```diff
@@ -0,0 +1,42 @@
+const path = require('path')
```
ai, line 1: Can you import methods like `let { resolve } = require(…)` to keep consistency with other files

packages/size-limit/read-pkg-up.js
4en1x: I see that configured version of node is `10^ || ...`, and `require('fs/promises')` are supported from version 12, and `require('fs').promises` are supported from version 11

ai: Got it 👍

ai (approved):

packages/size-limit/read-pkg-up.js
4en1x: Fixed in 84789586cb9d4aa93cb6945325a4d26933dce720

ai:
I will accept it when I will be at home

packages/size-limit/read-pkg-up.js
4en1x: Fixed in 84789586cb9d4aa93cb6945325a4d26933dce720

ai:
Released in 5.0

PR #236, opened 2021-06-24
Replace `ora` package with local spinner implementation

vadimyen:
I replaced `ora` according to the [task](https://cultofmartians.com/tasks/size-limit-dependencies.html#task) (btw, there's an incorrect link to the solution)


ai:
Thanks, I fixed the link.

I missed a critical feature of `ora`. Not every environment supports colors and animation. For instance, CI or `npx size-limit | grep limit` don’t support them.

ai:
Released in 5.0

PR #240, opened 2021-07-26
Add optional dependencies support

edouardmenayde:
Currently if you install size-limit presets in your as optional dependency (mainly useful not to download chromium if you don't need to in your CI using `--no-optional`) then size-limit will not find plugins.
This PR adds this feature.


ai:
Will it work if user will not install plugin from optional dependency?

edouardmenayde:
Yes it will, thanks for merging so quickly !

ai:
Thanks. Released in 5.0.2.

edouardmenayde:
You rock, thanks 🚀

PR #246, opened 2021-09-21
Update to webpack v5

### LINKED ISSUE #205: Update to webpack v5

ludofischer:
Since more projects like Next.js are moving to webpack 5 by default, I've had another go at https://github.com/ai/size-limit/issues/205. Moving to webpack 5 also helps with https://github.com/ai/size-limit/issues/232 since webpack 5 itself has fewer dependencies.

I've used the previous PR https://github.com/ai/size-limit/pull/206 as a basis.
I've struggled with updating the expected bundle sizes in tests as I could not understand exactly how the constants inn https://github.com/ai/size-limit/blob/86fe4662d0f90681127ec78759c720dab2f7c30a/packages/webpack/index.js#L12 were computed.
With webpack 5, the outputs seem to be sometimes larger and sometimes smaller.

Also I am not sure how to check that the bundle analyzer works correctly.
The previous blocker was webpack-bundle-analyzer not working with webpack 5 but it seems the most problematic issue has been solved in the mean time https://github.com/webpack-contrib/webpack-bundle-analyzer/pull/447

Fixes #205

### DIFF
--- a/packages/size-limit/test/get-config.test.js
+++ b/packages/size-limit/test/get-config.test.js
@@ -188,10 +188,8 @@ it('normalizes import', async () => {
     checks: [
       {
         import: {
-          [fixture(
-            'integration-esm',
-            'index.js'
-          )]: '{ VERY_LONG_NAME_FOR_CONST_TO_TEST_TREE_SHAKING }'
+          [fixture('integration-esm', 'index.js')]:
+            '{ VERY_LONG_NAME_FOR_CONST_TO_TEST_TREE_SHAKING }'
         },
         limit: '1 B',
         highlightLess: true,
--- a/packages/webpack/index.js
+++ b/packages/webpack/index.js
@@ -9,10 +9,10 @@ let convertConfig = require('./convert-config')
 let runWebpack = require('./run-webpack')
 let getConfig = require('./get-config')
 
-const WEBPACK_EMPTY_PROJECT = 962
-const WEBPACK_EMPTY_PROJECT_GZIP = 461
-const WEBPACK_EMPTY_PROJECT_IMPORT = 965
-const WEBPACK_EMPTY_PROJECT_IMPORT_GZIP = 473
+const WEBPACK_EMPTY_PROJECT = 0
+const WEBPACK_EMPTY_PROJECT_GZIP = 20
+const WEBPACK_EMPTY_PROJECT_IMPORT = 37
+const WEBPACK_EMPTY_PROJECT_IMPORT_GZIP = 57
 
 function getFiles(stats, check) {
   let entries = {}
@@ -30,11 +30,11 @@ function getFiles(stats, check) {
 
   return Object.keys(entries)
     .reduce((assets, i) => assets.concat(entries[i].assets), [])
-    .map(i => {
+    .map(({ name }) => {
       if (check.webpackConfig.output && check.webpackConfig.output.path) {
-        return join(check.webpackConfig.output.path, i)
+        return join(check.webpackConfig.output.path, name)
       } else {
-        return join(process.cwd(), 'dist', i)
+        return join(process.cwd(), 'dist', name)
       }
     })
 }
--- a/packages/size-limit/test/__snapshots__/run.test.js.snap
+++ b/packages/size-limit/test/__snapshots__/run.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`run allows to use peer dependencies in import 1`] = `
 "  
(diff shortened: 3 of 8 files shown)


ludofischer:
More info on issues with webpack-bundle-analyzer: https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/466 Apparently since webpack now aggressively concatenates modules by default, sometimes it is just impossible to show how much each module contributes to the output size, since the module boundaries are lost in the output.

packages/webpack/index.js
```diff
@@ -9,10 +9,10 @@ let convertConfig = require('./convert-config')
 let runWebpack = require('./run-webpack')
 let getConfig = require('./get-config')
 
-const WEBPACK_EMPTY_PROJECT = 962
-const WEBPACK_EMPTY_PROJECT_GZIP = 461
-const WEBPACK_EMPTY_PROJECT_IMPORT = 965
-const WEBPACK_EMPTY_PROJECT_IMPORT_GZIP = 473
+const WEBPACK_EMPTY_PROJECT = 0
```
ai, line 12: You need to fix these constants according to the new webpack loader in the bundle (code which webpack adds in additional to user’s code).

We are using them to subtract webpack’s own code from the bundle to calculate the cost of added library:

```
libraryCost = projectWithLib - emptyProject
```

Change the value to have `0` in empty project test https://github.com/ai/size-limit/blob/main/packages/size-limit/test/run.test.js#L300-L306

And change `IMPORT` constants to have `1` on importing the single digit https://github.com/ai/size-limit/blob/main/packages/webpack/test/index.test.js#L240-L258

ai:
> I've struggled with updating the expected bundle sizes in tests as I could not understand exactly how the constants inn 

You need to fix these constants according to the new webpack loader in the bundle (code which webpack adds in additional to user’s code).

We are using them to subtract webpack’s own code from the bundle to calculate the cost of added library:

```
libraryCost = projectWithLib - emptyProject
```

Change the value to have `0` in empty project test https://github.com/ai/size-limit/blob/main/packages/size-limit/test/run.test.js#L300-L306

And change `IMPORT` constants to have `1` on importing the single digit https://github.com/ai/size-limit/blob/main/packages/webpack/test/index.test.js#L240-L258

> Also I am not sure how to check that the bundle analyzer works correctly.

It is OK

> Also I am not sure how to check that the bundle analyzer works correctly.

If `--why` is showing Webpack Bundle Analyzer data it is OK. It is now Webpack Bundle Analyzer duty to find the best way to use webpack 5 API.

Anyway Webpack Bundle Analyzer data was never 100% accurate, it just help understand what happens, not find the exact numbers.

ludofischer:
OK I think I have updated the constants in `packages/webpack/index.js` to have the correct values, so the empty project has 0 bytes and the single import has 1B. There are still some discrepancies with the old numbers in other tests, I think that because webpack does not include any runtime at all for basic examples, the gzip size ends up larger than the non-gzip size.

packages/size-limit/test/get-config.test.js
```diff
@@ -188,10 +188,8 @@ it('normalizes import', async () => {
     checks: [
       {
         import: {
-          [fixture(
-            'integration-esm',
-            'index.js'
-          )]: '{ VERY_LONG_NAME_FOR_CONST_TO_TEST_TREE_SHAKING }'
+          [fixture('integration-esm', 'index.js')]:
```
ludofischer, line 191: I have touched this file once and the git hook keeps wanting to reformat it although there are no changes.

ai:
Looks great. Give me a few days, I need to release one fix in patch release of Size Limit, and then I will start preparing major release.

PR #259, opened 2021-10-11
Save report into tmp dir by default

smelukov:
Size-limit clears output directory by default (with statoscope report).
By this PR statoscope will save a report into os temp directory.
Also, we can pass saveBundle directory, and then a report will be saved there


smelukov:
Did you link fresh size-limit for nanoid?

Issue #267, opened 2021-10-29
Add `@size-limit/esbuild` plugin


ai:
Moving to esbuild will reduce `node_modules` size and improve performance

Windrushfarer:
Hey, I found this issue in `cultofmartians` :) I will take a look at it, doesn't look too complicated

ai:
Yeap. We need to cover the simple case of libraries like `nanoid` (when we have multiple JS files with `import` or `require()` and `process.env.NODE_ENV`). No CSS or any other webapp bundling problems.

PR #270, opened 2021-11-14
Esbuild plugin

Toomean:
Hello there! I've worked on a plugin for esbuild( according to #267 ).
Not sure about the needed requirements and the final point for this

Take a look, please


packages/esbuild/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@size-limit/esbuild",
+  "version": "6.0.4",
+  "description": "Esbuild plugin for Size Limit",
```
ai, line 4:
```suggestion
  "description": "esbuild plugin for Size Limit",
```

Let’s use downcased name. esbuild’s author prefer it.

ai:
Looks excellent. We just need also `check.import` support to test libraries with tree-shaking.

Search for `check.import` and `import` in the project and you will find a code in `webpack` plugin (and config validation in `size-limit`) which you will need to back port (copy-pasting is OK, but if it will be possible to move some helpers to `size-limit` and then re-use it, it will better).

packages/esbuild/README.md
```diff
@@ -0,0 +1,12 @@
+# Size Limit Esbuild Plugin
+
+The plugin for [Size Limit] to track dependencies size by Esbuild.
```
ai, line 3: We need to fix `Esbuild` → `esbuild` across all docs, not only in the place where I showed

packages/esbuild/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "engines": {
+    "node": "^12.0.0 || ^14.0.0 || >=16.0.0"
+  },
+  "peerDependencies": {
+    "size-limit": "6.0.4"
+  },
+  "dependencies": {
+    "@statoscope/webpack-plugin": "^5.17.0",
+    "css-loader": "^6.5.0",
+    "css-minimizer-webpack-plugin": "^3.0.2",
```
ai, line 25: Why do we need all these dependencies?

packages/esbuild/index.js
```diff
@@ -0,0 +1,117 @@
...
+let { nanoid } = require('nanoid/non-secure')
+let { tmpdir } = require('os')
+let { join, resolve, parse } = require('path')
+let rm = require('size-limit/rm')
+
+let convertConfig = require('./convert-config')
+let runEsbuild = require('./run-esbuild')
+let getConfig = require('./get-config')
+
+const ESBUILD_EMPTY_PROJECT = 0
```
ai, line 12: We will need tests with empty project to validate these constants.

The project with empty file should have `0` size.

The project with one digit export (and `import` config) should have `1` size.

packages/webpack/get-config.js
```diff
@@ -1,27 +1,14 @@
 let StatoscopeWebpackPlugin = require('@statoscope/webpack-plugin').default
-let { writeFile } = require('fs').promises
+let updateCheckWithImports = require('size-limit/updateCheckWithImports')
```
ai, line 2: I do not use camel case in other files in the project. Be careful.

packages/esbuild/package.json
Toomean: nope, I'm going to remove them

packages/esbuild/README.md
Toomean: Changed, I've set `"(by |)esbuild",` for the yaspeller dictionary. How can it be properly handled?

packages/esbuild/package.json
Toomean: Removed

packages/webpack/get-config.js
Toomean: Thanks for the point! I will keep an eye on it

ai:
We still need to fix coverage and fix some strange Node.js 12 issue

packages/esbuild/index.js
Toomean: I'm struggling with this a bit. Could you help me with that? Maybe we have an example of how to validate such kinds of constants

ai: Here is a test example for `*_EMPTY_PROJECT_IMPORT` and `*_EMPTY_PROJECT_IMPORT_GZIP`
https://github.com/ai/size-limit/blob/main/packages/webpack/test/index.test.js#L241-L268

Here is a test example for `*_EMPTY_PROJECT` and `*_EMPTY_PROJECT_GZIP` (yeap, it is in `size-limit` and not in `webpack` plugin for historical reason, feel free to move).
https://github.com/ai/size-limit/blob/main/packages/size-limit/test/run.test.js#L322-L328

ai:
Test coverage shows that we need to calibrate (test) `ESBUILD_EMPTY_PROJECT_IMPORT` and `ESBUILD_EMPTY_PROJECT_IMPORT_GZIP`.

Also, we still have Node.js 12 problem:

```
> error: Could not resolve "/home/runner/work/size-limit/size-limit/packages/esbuild/test/fixtures/unknown.js"

 > error: Refusing to overwrite input file "../../../../../tmp/size-limit-qzG5j_Arw23-2HlwpuSXJ/index.js" (use "allowOverwrite: true" to allow this)

 > error: Refusing to overwrite input file "../../../../../tmp/size-limit-8Rz_8m1YKc_k-mX_IJzsY/index.js" (use "allowOverwrite: true" to allow this)
```

Toomean:
Can't reproduce errors from actions on my local machine, will investigate it on another os. As I see, this problem appears on every node version

ai:
> Can't reproduce errors from actions on my local machine

It could be related to CI environment also

ai:
I pushed merge commit to run CI

ai:
Greta work. Thanks.

PR #271, opened 2021-11-15
Added @size-limit/webpack-why and @size-limit/webpack-css packages

### LINKED ISSUE #266: Reduce `@size-limit/webpack` size
Moved Statoscope and CSS dependencies to `@size-limit/webpack-why` and `@size-limit/webpack-css` will reduce package size a lot

Windrushfarer:
Fixes #266

I moved dependecies for analyzing and CSS support into separate packages. The only question that I am not sure is about adding these plugins to any of the presets. `Webpack` plugin had all its features by default, and now it's not.

I also had to update snapshots and some of the tests from `size-limit` package.

Btw it seems like a big PR and if you find it tough to merge, we can split it into 2, so 1 PR per new package.


ai:
One PR is good.

1. Can you respect 80 columns line? You can install Prettier to text editor, it will format code for you.
2. We need to fix code coverage.

Windrushfarer:
@ai thanks for your review :) I pushed new changes

PR #272, opened 2021-11-25
docs: fix action version in workflow example

Drapegnik:
to use latest version, `v1.0.0` is outdated

https://github.com/andresz1/size-limit-action/issues/68#issuecomment-979479985


ai:
Thanks

PR #288, opened 2022-07-06
feat: add `sizeLimit` to json output

elliotwestlake:
Add `sizeLimit` to `json` output


ai:
Don’t worry. CI error is from Jest and ESM and is already in `main`

ai:
Thanks for the wait. We had a problem with CI.

The fix was released in 8.0.

PR #303, opened 2022-12-06
chore: update and fix pnpm version

unional:
The version of pnpm I have on my wsl was old and I notice it generates different lock file.

Updating package.json to indicate this repo is using pnpm and update the version.

If you want to pin it to another version, that works too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,5 +84,6 @@
     "semi": false,
     "singleQuote": true,
     "trailingComma": "none"
-  }
+  },
+  "packageManager": "pnpm@7.18.0"
 }


unional:
Another thing I noticed is that there are some cyclic dependencies:

```sh
pnpm i
Scope: all 12 workspace projects
 WARN  There are cyclic workspace dependencies
Lockfile is up to date, resolution step is skipped
Already up to date
```

Is it something you plan to fix or you think it can be kept as is?

ai:
I don’t want to update pnpm version every time I am changing it (I have too many projects).

What we can do?

unional:
> I don’t want to update pnpm version every time I am changing it (I have too many projects).

This will fix it to this particular version (7.18.0)
For myself, I use mergify + renovate + dependabot so that those dependencies are updated automatically.

ai:
> For myself, I use mergify + renovate + dependabot so that those dependencies are updated automatically.

I have 20+ projects. PRs from these services will ruin my day.

Can we use non-specific version there? Something like `pnpm@^7.18.0`

unional:
https://nodejs.org/api/packages.html#packagemanager
Seems like it does not.

We can drop this if you don't want. Most of the time it will be fine.

For myself, this ensure we have the same development environment just like the benefits of having a lock file.

> I have 20+ projects. PRs from these services will ruin my day.

Yeah, I can feel you. I have 100+ 😛

PR #307, opened 2023-01-10
chore(deps): bump devDependencies

piwysocki:


ai:
CI fails.

For test (which fails because of small size difference) it is safe just to change the size a little.

PR #308, opened 2023-01-18
fix: prevent spinner output while --silent

isuvorov:
Prevent messages like: `✔ Adding to empty esbuild project` while enabled silent mode


ai:
Thanks. Released in 8.1.2.

PR #312, opened 2023-02-12
feat(@size-limit/esbuild-why): support multiple sections for `--why`

### LINKED ISSUE #311: multiple checks with `--why` only kept last `report.html`
e.g.:

```
// package.json
{
  "size-limit": [
    {
      "path": "./cjs/index.js",
      "limit": "15 kb"
    },
    {
      "path": "./esm/index.js",
      "limit": "15 kb"
    },
    {
      "path": "./tslib/index.js",
      "limit": "15 kb"
    }
  ]
}
```

Running `size-limit --why` (using `esbuild` and `esbuild-why`) will create a `report.html` containing the result of the last one (`tslib`) as the previous ones are overwritten.

I can change it to add `report-{x}.html`, but a better way is giving each check a name, and name the output as `report-{name}.html`?

Currently the `check` contains a `name`, but it is the entry file name:

```
{
  path: './tslib/index.js',
  limit: '15 kb',
  files: [ 'D:\\code\\unional\\type-plus\\tslib\\index.js' ],
  name: './tslib/index.js',
  sizeLimit: 15000,
  ...
}
```

unional:
Fixed #311

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,7 +577,8 @@ inserts `style-loader` runtime (≈2 kB) into the bundle.
 
 You can run `size-limit --why` to analyze the bundle.
 
-You will need to install `@size-limit/esbuild-why` or `@size-limit/webpack-why` depends on which bundler you are using (default is `esbuild`).
+You will need to install `@size-limit/esbuild-why` or `@size-limit/webpack-why`
+depends on which bundler you are using (default is `esbuild`).
 
 For `@size-limit/esbuild-why`,
 it will generate a `esbuild-why.html` at the current directory.
--- a/README.md
+++ b/README.md
@@ -573,6 +573,39 @@ inserts `style-loader` runtime (≈2 kB) into the bundle.
 [Statoscope docs]: https://github.com/statoscope/statoscope/tree/master/packages/webpack-plugin#optionsreports-report
 [pattern]: https://github.com/sindresorhus/globby#globbing-patterns
 
+## Analyze with `--why`
+
+You can run `size-limit --why` to analyze the bundle.
+
+You will need to install `@size-limit/esbuild-why` or `@size-limit/webpack-why` depends on which bundler you are using (default is `esbuild`).
+
+For `@size-limit/esbuild-why`,
+it will generate a `esbuild-why.html` at the current directory.
+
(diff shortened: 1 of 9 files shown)


packages/esbuild-why/index.js
```diff
@@ -1,14 +1,16 @@
-let { visualizer } = require('esbuild-visualizer')
-let { join } = require('path')
-let { writeFileSync } = require('fs')
+const { visualizer } = require('esbuild-visualizer')
```
ai, line 1: I use `let` :)

packages/esbuild-why/report.js
```diff
@@ -0,0 +1,10 @@
+const isValidFilename = require('valid-filename')
```
ai, line 1: The check is too simple to use another dependency. Let’s just copy regexp

https://github.com/sindresorhus/filename-reserved-regex/blob/main/index.js

README.md
```diff
@@ -573,6 +573,39 @@ inserts `style-loader` runtime (≈2 kB) into the bundle.
 [Statoscope docs]: https://github.com/statoscope/statoscope/tree/master/packages/webpack-plugin#optionsreports-report
 [pattern]: https://github.com/sindresorhus/globby#globbing-patterns
 
+## Analyze with `--why`
+
+You can run `size-limit --why` to analyze the bundle.
+
+You will need to install `@size-limit/esbuild-why` or `@size-limit/webpack-why` depends on which bundler you are using (default is `esbuild`).
```
ai, line 580: We need to keep 80 columns line

packages/esbuild-why/index.js
unional: Oh, also at the top-level?

I looked up the `eslint-plugin-prefer-let` it says you can use `const` at top-level. 😛 

I can change that.

ai: I use `const` only for cases like `const MAX = 100`

packages/esbuild-why/report.js
unional: I pulled in the code directly. Added attribution

ai:
Thanks. Released in 8.2.4.

Issue #326, opened 2023-06-09
Loosen strictness of peer dependency requirement of `tslib`


mcmxcdev:
See more: https://github.com/LironHazan/nx-size-limit/issues/30

Seems its coming from https://github.com/ai/size-limit/blob/e52d69af19ffee2216dcdeb8ec77361a26916f5d/pnpm-lock.yaml#L4873-L4882

ai:
Sure, send PR

PR #327, opened 2023-06-09
chore: loose strictness of peer dependency requirement of `tslib`

mcmxcdev:
Still no clue where the peer dependency requirement comes from, I don't see it defined in either `size-limit-node-esbuild` or `size-limit` itself.

Now users could update to `tslib` 2.5.3 at least to satisfy and match the dependency requirement.

Closes https://github.com/ai/size-limit/issues/326

***

- re-install `size-limit-node-esbuild` to refresh `tslib` from 2.5.0 to 2.5.3

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4876,7 +4876,7 @@ packages:
     dependencies:
       '@size-limit/esbuild': 8.2.4(size-limit@packages+size-limit)
       nanoid: 3.3.6
-      tslib: 2.5.0
+      tslib: 2.5.3
     transitivePeerDependencies:
       - size-limit
     dev: true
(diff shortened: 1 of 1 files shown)


ai:
It comes from `package.json` (in project’s root) and `fixtures/plugins/package.json`

mcmxcdev:
> It comes from `package.json` (in project’s root) and `fixtures/plugins/package.json`

That's where `size-limit-node-esbuild` is installed yes, but not where `peerDependencies: { "tslib": "2.5.0" }` is coming from.

I think I figured out what the actual issue is, read more: https://github.com/ai/size-limit/pull/327#issuecomment-1585048284

ai:
Have no idea. The lock file which you fixed is not in any npm package.

PR #328, opened 2023-06-23
@size-limit/esbuild-why: Automatically open esbuild reports 

slavik-chapelskyi:
This small change remove the need to manually open generated esbuild reports


ai:
Thanks. Released in 8.2.5.

Issue #331, opened 2023-07-24
React should be peer dependency


ad1992:
In the package `@size-limit/time`, `react @17.x` is a dependency which leads to installing multiple versions in host and increasing bundle size as well. 

https://github.com/ai/size-limit/blob/main/packages/time/package.json#L24

It should be instead a peer dependency where 17.x / 18.x should be allowed whichever host is using.

ai:
`@size-limit/time` is not using on client-side. It is server-side only package.

Are you sure it leads to increasing bundle size? Can you show some project?

ad1992:
> `@size-limit/time` is not using on client-side. It is server-side only package.
> 
> Are you sure it leads to increasing bundle size? Can you show some project with this issue?

Since its a dev dependency so bundle size eventually won't matter but due to multiple versions of react the host build can break if they are on React 18 specially thats what is happening right now https://github.com/excalidraw/excalidraw/issues/6801

ai:
Hm. The problem is that we can’t use host’s version.

We are using React sources to estimate to CPU power of CI by [parsing these sources](https://github.com/ai/size-limit/blob/main/packages/time/get-running-time.js#L5-L6). React 18 and React 17 has a different size and the result will be different.

But we can switch from React to some another library. Do you know some big library without dependenies?

ad1992:
> parsing these sources

In that case instead of using the npm package for `react`, you can keep a local copy of its `react.production.min.js` ? or even use the [unpkg link](https://unpkg.com/browse/react@17.0.2/umd/react.production.min.js) ? 
That way host don't have to install `react`

ai:
Hm, interesting idea. Do you want to send PR to keep your name in the project’s history?

ad1992:
Sure I will raise a PR shortly

ad1992:
@ai had been busy so couldn't raise the PR, I have opened a [PR](https://github.com/ai/size-limit/pull/334) and this seems to work fine, let me know

PR #334, opened 2023-09-04
fix: remove react dependency and use unpkg link instead

ad1992:
fixes https://github.com/ai/size-limit/issues/331


ai:
Thanks. Released in 9.0.

PR #338, opened 2023-10-22
Move to ESM

lev875:


packages/esbuild-why/test/index.test.js
```diff
@@ -1,12 +1,15 @@
...
-let [esbuild] = require('@size-limit/esbuild')
-let {join} = require('path')
-let rm = require('size-limit/rm')
-let open = require('open')
+import esbuildPkg from '@size-limit/esbuild'
+import {readFile} from 'fs/promises'
+import open from 'open'
+import {join} from 'path'
+import rm from 'size-limit/rm'
+import { afterEach, expect, it, vi } from "vitest"
```
ai, line 6: Can you run Prettier on changed files?

I have it in VS Code. But you can install to the project, run, and uninstall.

packages/esbuild/test/index.test.js
```diff
@@ -63,64 +66,103 @@ it('uses esbuild to make bundle', async () => {
...
+      checks: [{ config: fixture('esm/esbuild.config.js'), entry: ['small'] }],
+      configPath: ROOT_CONFIG
+    }
+    await run(config)
+    expect(config.checks[0].size).toBe(68)
+  })
+})
+
+describe('throws error on unknown entry', () => {
+  it('should work with commonjs config', async () => {
```
ai, line 116:
```suggestion
  it('works with commonjs config', async () => {
```

and all other tests

packages/esbuild-why/test/index.test.js
lev875: Ah, damn, I did `eslint --fix` but forgot about prettier. I've installed it globally and ran `pnpm exec prettier **/*.js --write`, should do the trick.

packages/esbuild/test/index.test.js
lev875: Fixed by [c17c150](https://github.com/ai/size-limit/pull/338/commits/c17c1507498d08ff3aa9f05cd8f20b02843f99ff)

PR #345, opened 2023-11-10
fix: use dynamic `import` directly without `require.resolve`

### LINKED ISSUE #336: Unable to run on windows 10
Hi

I'm using size-limit 9 with NodeJs 18 and windows 10 

Using powershell I have de following error.

```powershell
C:\desarrollos\ActiveDirectory>npm run size --why
npm ERR! No workspaces found:
npm ERR!   --workspace=--usage
```

The second option y run at WSL but I have one annoying bug that I can solve

```
⠋ Running JS in headless ChromeError: Failed to launch the browser process! spawn C:\Program Files\Google\Chrome\Application\chrome.exe ENOENT
```

It's possible to run `npm run size --why` and saving the result, without open the browser?

### LINKED ISSUE #344: plugins are `import`ed with `require.resolve` which will resolve `commonjs` entry for dual package
https://github.com/ai/size-limit/blob/f2493379fefc00559af2a8c54fc52cedbbc9b0bd/packages/size-limit/load-plugins.js#L30-L34

This is unexpected for third-party plugins.

JounQin:
close #336
close #344

### DIFF
--- a/packages/size-limit/load-plugins.js
+++ b/packages/size-limit/load-plugins.js
@@ -1,6 +1,3 @@
-import { createRequire } from 'node:module'
-const require = createRequire(import.meta.url)
-
 function toArray(obj) {
   return typeof obj === 'object' ? Object.keys(obj) : []
 }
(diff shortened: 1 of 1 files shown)


JounQin:
image

Some entries are not exported, should we export then? Or we still use the absolute path hack?

@ai

ai:
Yes, add missed exports to `index.js` export of the package.

ai:
> Some entries are not exported, should we export then?

I think we should not need `@size-limit/esbuild/convert-config` export at all. This is very simple function. Just replace it.

ai:
If you need some bigger function, I can export it.

ai:
Thanks. Released in 10.0.3.

JounQin:
Ah, I just realized that was my plugin throwing, actually the newer versions are not using the unexported entries now. So we'd better upgrade the test case dep versions.

Issue #343, opened 2023-11-09
Switch from GZIP to Brotli by default


kytta:
Brotli, as we know, is a more superior compression algorithm; fast & effective. According to Can I Use, [97% of devices can use Brotli](https://caniuse.com/brotli). It's widely adopted by the servers, too: nginx, Apache, LiteSpeed, lighttpd, Caddy, and others. Big CDNs like [Fastly](https://www.fastly.com/blog/brotli-is-great-at-making-things-small-and-no-we-arent-talking-about-the) and [Cloudflare](https://blog.cloudflare.com/this-is-brotli-from-origin/) are enabling it by default.

Wouldn't it make sense for size-limit to flip the switch, too, and show the "brotlied" result by default, making GZIP opt-in and encouraging people to switch their infrastructure?

ai:
Sounds like a good idea. But I will ask in Twitter first to collect the feedback.

ai:
Here is PR https://github.com/ai/size-limit/pull/346

ai:
Here is a good statistics about the real web traffic in 2022 https://almanac.httparchive.org/en/2022/cdn#brotli-adoption

* 47% of CDN traffic is Brotli
* 25% of self-hosted servers

I think it is pretty solid numbers. In 2023, they will be better. And this is the whole traffic. Size Limit users is a little more advanced.

![Screenshot from 2023-11-11 07-11-26](https://github.com/ai/size-limit/assets/19343/45e26175-4c44-4548-90c9-f92a251d48e9)

ai:
Released in 11.0.

PR #346, opened 2023-11-11
Move to Brotli by default

ai:
Fixes https://github.com/ai/size-limit/issues/343


fixtures/peer-esbuild-gzip/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "peer-esbuild-non-gzip",
```
kytta, line 2: I think this line should be renamed, too. As far as I can see on my small phone screen, this applies to every package.json that was moved, but where the package was not renamed. Doesn't make a difference, but it'd be more correct to rename them

kytta (changes requested):
Looks good to me, but as mentioned in the comment above, I would also change the `name` property of the fixtures' `package.json`s

fixtures/peer-esbuild-gzip/package.json
ai: I just removed most of `package.name` since many of them was outdated baba6aa

kytta (approved):

PR #357, opened 2024-03-12
Fix StatoscopeWebpackPlugin import to support default export

hoo00nn:
- [x] Fixes the `TypeError: StatoscopeWebpackPlugin is not a constructor` error.

An error `TypeError: StatoscopeWebpackPlugin is not a constructor` occurs if the module does not export as default.

@ai Hello, I wanted to bring to your attention an issue I've found while utilizing the `@size-limit/webpack-why` package. Since I got an error message (`StatoscopeWebpackPlugin is not a constructor`), I have consulted with the [official documentation](https://github.com/statoscope/statoscope/tree/master/packages/webpack-plugin). They advised that the class should be imported using the default export.

I have submitted a PR addressing this matter and would greatly appreciate it if you could take a moment to review the proposed changes. Your feedback would be invaluable.

### DIFF
--- a/packages/webpack-why/index.js
+++ b/packages/webpack-why/index.js
@@ -6,7 +6,7 @@ function addStatoscope(limitConfig, check, webpackConfig) {
     let shouldOpen = process.env.NODE_ENV !== 'test' && !limitConfig.saveBundle
 
     webpackConfig.plugins.push(
-      new StatoscopeWebpackPlugin({
+      new StatoscopeWebpackPlugin.default({
         additionalStats: [limitConfig.compareWith, check.compareWith].filter(
           Boolean
         ),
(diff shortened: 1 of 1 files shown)


ai:
Strange that we do not catch it with tests

ai:
Now we have a broken CI since we tested it
https://github.com/ai/size-limit/actions/runs/8250644947/job/22565763246

Something strange happens. Can you investigate and send another PR (I will revert this one)?

hoo00nn:
> Now we have a broken CI since we tested it
https://github.com/ai/size-limit/actions/runs/8250644947/job/22565763246
> Something strange happens. Can you investigate and send another PR (I will revert this one)?


@ai Yes, I'll check.

hoo00nn:
@ai, Hello

After further investigation, I was unable to pinpoint the exact cause of the issue. However, I've observed that depending on the project, the module in question may operate in either ES or CommonJS mode. To prevent any issues between the two module systems, I've made the necessary code fix and submitted the PR. I would appreciate your feedback on this.

all tests are passing.

- https://github.com/ai/size-limit/pull/358

PR #358, opened 2024-03-12
Fix ensure compatibility with ES and CommonJS module systems for StatoscopeWebpackPlugin

hoo00nn:
- [x] Fix compatibility issues arising from differences between CommonJS and ES module systems.

### DIFF
--- a/packages/webpack-why/index.js
+++ b/packages/webpack-why/index.js
@@ -2,7 +2,7 @@ import _StatoscopeWebpackPlugin from '@statoscope/webpack-plugin'
 import { join } from 'node:path'
 
 function addStatoscope(limitConfig, check, webpackConfig) {
-  const StatoscopeWebpackPlugin = _StatoscopeWebpackPlugin.default ?? _StatoscopeWebpackPlugin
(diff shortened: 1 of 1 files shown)


packages/webpack-why/index.js
```diff
@@ -1,7 +1,8 @@
-import StatoscopeWebpackPlugin from '@statoscope/webpack-plugin'
+import _StatoscopeWebpackPlugin from '@statoscope/webpack-plugin'
 import { join } from 'node:path'
 
 function addStatoscope(limitConfig, check, webpackConfig) {
+  let StatoscopeWebpackPlugin = _StatoscopeWebpackPlugin.default ? _StatoscopeWebpackPlugin.default : _StatoscopeWebpackPlugin;
```
ai, line 5: Do you find a reason why it works sometime with `default` sometimes not?

hoo00nn: I also don't know why it behaves that way. However, in my case, the module operates using the CommonJS format. Due to that error, I have to remove the `--why` option and configure it manually. What do you think about supporting this?

I've created a [test project](https://github.com/hoo00nn/size-limit-test), and the same issue occurs. How is it in your environment?

ai:
Thanks. Released in 11.1.

PR #359, opened 2024-03-16
Fix config file path resolution on windows

aryaemami59:
## This PR:

  - [X] Fixes config file path resolution on windows. On windows sometimes you could get `Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]` which should now be fixed due to usage of `pathToFileURL`.


aryaemami59:
For what it's worth I think I should probably try and make the tests OS agnostic so we can check for potentital issues like this during CI.

ai:
We can add Windows to CI

aryaemami59:
@ai Yeah but the snapshots make the tests fail on windows.

PR #361, opened 2024-03-21
Support css content in esbuild plugin

### LINKED ISSUE #360: Esbuild plugin produces incorrect numbers when CSS used
If a project contains a CSS file in imports, for example `import "./styles.css";`, size-limit with esbuild plugin only counts CSS size and skips Javascript. The same works well on webpack + webpack-css plugin.

Reproducible in this sample project: https://github.com/just-boris/size-limit-demos

The same entry produces different results when using esbuild and webpack plugins

just-boris:
`parse('index.css').name` and `parse('index.js').name` prodiuces the same `index` key, which overwrites one of the entries. 

Use `base` to prevent conflicts and use `index.js`/`index.css` as the keys.

Fixes #360

### DIFF
--- a/packages/esbuild/index.js
+++ b/packages/esbuild/index.js
@@ -23,18 +23,20 @@ function getFiles(buildResult, check) {
   let outputs = buildResult.metafile.outputs
 
   for (let key in outputs) {
-    let outputEntryName = parse(key).name
+    let outputEntryName = parse(key).base
     outputs[outputEntryName] = outputs[key]
     outputs[outputEntryName].path = resolve(key)
     delete outputs[key]
   }
 
   if (check.entry) {
-    for (let i of check.entry) {
-      if (outputs[i]) {
-        entries[i] = outputs[i]
-      } else {
-        throw new SizeLimitError('unknownEntry', i)
+    for (let entry of check.entry) {
+      let matches = Object.keys(outputs).filter(key => parse(key).name === entry)
+      if(matches.length === 0) {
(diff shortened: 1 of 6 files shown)


ai:
Lgtm. Do you want to add anything (PR is marked as draft) or we ready for merge?

packages/esbuild/index.js
```diff
@@ -23,18 +23,20 @@ function getFiles(buildResult, check) {
...
   }
 
   if (check.entry) {
-    for (let i of check.entry) {
-      if (outputs[i]) {
-        entries[i] = outputs[i]
-      } else {
-        throw new SizeLimitError('unknownEntry', i)
+    for (let entry of check.entry) {
+      let matches = Object.keys(outputs).filter(key => parse(key).name === entry)
```
just-boris, line 34: Make sure entry matching still uses `path.parse(entry).name` as before

packages/size-limit/size-limit-error.js
```diff
@@ -37,7 +37,7 @@ const MESSAGES = {
   unknownArg: arg =>
     `Unknown argument *${arg}*. Check command for typo and read docs.`,
   unknownEntry: entry =>
-    `Size Limit didn’t find *${entry}* entry in custom Webpack config`,
+    `Size Limit didn’t find *${entry}* entry in the custom bundler config`,
```
just-boris, line 40: Spent some time in confusion how esbuild change may break anything with webpack config. Let's rename the error message to a more generic

PR #365, opened 2024-06-06
Add explicit `packageManager` field to `package.json`

aryaemami59:
## **This PR:**

- [X] Adds explicit `packageManager` field to `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,5 +35,6 @@
     "semi": false,
     "singleQuote": true,
     "trailingComma": "none"
-  }
+  },
+  "packageManager": "pnpm@9.2.0"
 }


ai:
I have 50+ open source projects, it will be very hard for me keep up to date to version of pnpm in package.json manually in every project.

Do you have any reason why this key is important to spend more time on maintenance?

ai:
I really afraid of locking the exact version of pnpm.

What if there is old project with 9.1.1 and new developer come with 9.2? Will it require too many versions of pnpm on developers machine in corepack?

aryaemami59:
In that case, they will already have version 9.1.1 on their machine, since corepack caches the old versions on the global registry, and will not remove them unless they call `corepack cache clear` manually.

ai:
I decided to not use it right now. It is better to use the latest pnpm.

PR #368, opened 2024-06-22
chore: Ant Design is using Size Limit

afc163:
https://github.com/ant-design/ant-design/pull/49536#issuecomment-2181973455

图片


ai:
Thanks! Nice to see

PR #383, opened 2025-01-28
fix: remove extra blank line in terminal when output is empty

### LINKED ISSUE #363: Extra blank lines when using --silent
Extra blank lines appear when using the `--silent` option.

I propose add a check if (isSilentMode) in the appropriate parts of the code to fix this issue.

```ts
function createHumanReporter(process, isSilentMode = false) {
  function print(...lines) {
    if (!isSilentMode) {
      process.stdout.write('  ' + lines.join('\n  ') + '\n')
    }
  }

```
https://github.com/ai/size-limit/blob/d0d681d7d0c686773650ba1af03d892f0081e761/packages/size-limit/create-reporter.js#L48

Need advice on the correctness of this approach.

I could make PR by myself.

This gap is too БЕСИТ 😬
 
image
image
image

EvgenyWas:
The simpler solution is to use `process.stdout.moveCursor()` and `process.stdout.clearLine()` directly. However, i[f TTY is not available we can not use those methods](https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000645510/comments/115000530624) and consuming `readline` is a better-supported solution to do that.

Resolves #363

### DIFF
--- a/packages/size-limit/create-reporter.js
+++ b/packages/size-limit/create-reporter.js
@@ -1,5 +1,6 @@
 import bytes from 'bytes-iec'
 import { join } from 'node:path'
+import readline from 'node:readline'
 import pc from 'picocolors'
 
 const { bgGreen, bgRed, black, bold, gray, green, red, yellow } = pc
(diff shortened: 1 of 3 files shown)


ai:
Am I right than in non-TTY env it will do nothing (double line, which is OK)?

EvgenyWas:
> Am I right than in non-TTY env it will do nothing (double line, which is OK)?

No, it will work as expected. I mentioned TTY here just because `process.stdout.moveCursor()` was failing with the following error `TypeError: process.stdout.moveCursor is not a function` because the process is being run out of a TTY context during running tests. The usage of readline is more reliable here.

PR #386, opened 2025-03-21
Add documentation for `--save-bundle` flag

just-boris:
I keep forgetting the name of this flag, can we print it in the CLI help?


ai:
Nice, thanks. 

Can't you just update snapshots too by running `pnpm vitest run -u`?

ai:
Thanks. I will release with some feature because making release is a little hard right now.

Issue #387, opened 2025-04-07
Transitive Dependency tar-fs resolves to vulnerable version


GaryBarnes17:
## Description
- [More info on the vulnerability here](https://security.snyk.io/vuln/SNYK-JS-TARFS-9535930). 
- Basically, the `tar-fs` transitive dependency resolves to a vulnerable version (v3.0.5). 
- This should be upgraded to a version where this vulnerability is patched (v3.0.8)
- To see this, run `pnpm why tar-fs`. The output should look something like this:

```
devDependencies:
@size-limit/preset-app link:packages/preset-app
└─┬ @size-limit/time link:packages/time
  └─┬ estimo 3.0.3
    ├─┬ find-chrome-bin 2.0.2
    │ └─┬ @puppeteer/browsers 2.7.1
    │   └── tar-fs 3.0.8
    └─┬ puppeteer-core 22.6.5
      └─┬ @puppeteer/browsers 2.2.2
        **└── tar-fs 3.0.5**
@size-limit/preset-big-lib link:packages/preset-big-lib
└─┬ @size-limit/time link:packages/time
  └─┬ estimo 3.0.3
    ├─┬ find-chrome-bin 2.0.2
    │ └─┬ @puppeteer/browsers 2.7.1
    │   └── tar-fs 3.0.8
    └─┬ puppeteer-core 22.6.5
      └─┬ @puppeteer/browsers 2.2.2
        **└── tar-fs 3.0.5**
@size-limit/time link:packages/time
└─┬ estimo 3.0.3
  ├─┬ find-chrome-bin 2.0.2
  │ └─┬ @puppeteer/browsers 2.7.1
  │   └── tar-fs 3.0.8
  └─┬ puppeteer-core 22.6.5
    └─┬ @puppeteer/browsers 2.2.2
      **└── tar-fs 3.0.5**
```

GaryBarnes17:
> Can you ask `estimo` to update?

Yep, I absolutely can!

PR #388, opened 2025-04-07
fix: bump tar-fs to resolve vulnerability

GaryBarnes17:
## What it does 
- Resolves https://github.com/ai/size-limit/issues/387
- Upgrades the `tar-fs` transitive dependency to a non-vulnerable version via pnpm overrides. 
- Open to any feedback here - I can scope this override to specific package(s) if needed 

## How to Test
- run `pnpn why tar-fs`, you should see that all versions of this package resolve to v3.0.8

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,5 +35,10 @@
     "semi": false,
     "singleQuote": true,
     "trailingComma": "none"
+  },
+  "pnpm": {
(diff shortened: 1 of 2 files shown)


ai:
How it will solve the problem for non-pnpm size-limit users?

PR #389, opened 2025-04-28
feat: support zstd

SevereCloud:
Add support zstd

- [Node.js v22.15.0](https://nodejs.org/en/blog/release/v22.15.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,7 @@ For example, if you want to use `@size-limit/webpack`, you can just use
 
 Plugins:
 
-* `@size-limit/file` checks the size of files with Brotli (default), Gzip
+* `@size-limit/file` checks the size of files with Brotli (default), Zstd, Gzip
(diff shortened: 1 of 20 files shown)


ai:
I found that zstd is worse than brotli and the main reason to use it for dynamic HTML.

So you have any reason to use zstd?

SevereCloud:
> So you have any reason to use zstd?

Compression / decompression speed

One way or another, size-limit should help in determining the usefulness of zstd

ai:
> Compression / decompression speed

But what is the reason to use zstd for static assets (like JS bundle file which we often test in Size Limit)?

ai:
> Measuring the size using size-limit can help determine the usefulness of zstd

Releasing and support a feature is a lot of work for benchmark. It is better for you to do it with a fork and then report the usefulness :D

PR #392, opened 2025-07-29
fix(deps): upgrade estimo in time package

jonathanlee46:
This upgrades the `estimo` package in `packages/time`, which in turn upgrades the version of `puppeteer/packages/browsers`, which in turn upgrades `tar-fs` out of [this vulnerability range](https://github.com/advisories/GHSA-pq67-2wwv-3xjx).

### DIFF
--- a/packages/time/package.json
+++ b/packages/time/package.json
@@ -25,7 +25,7 @@
     "size-limit": "11.2.0"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


ai:
You don’t need to do it since `^` in version requirement means any `3.x.x` version. Just update nested dependencies by `npm audit --fix` or any nested dependencies update way.

PR #397, opened 2026-04-09
Add disablePlugins as option

### LINKED ISSUE #396: Proposal having two bundlers in one .size-limit.json
First thanks a lot for the great tool. 

I do have a feature request (couldn't find similar issues on this). It might be a little tough, as I saw that the plugins are auto-loaded.

## Status quo

Now when having a `.size-limit.json` with couple of configs then it will use the bundler which is installed, e.g. `@size-limit/webpack`. This works fine when all configs need to be build with webpack.

However, when adding `@size-limit/esbuild` as well, then everything will be bundled with esbuild and there is no way of changing that. Also the `webpack: boolean`, config is slightly confusing when there is the option to also have esbuild. It is actually also stated that `esbuild` is the default bundler, and that is ok.

## Proposal

It'd be nice, when both bundlers are installed, that it is still possible to choose which config uses which bundler. E.g. having this as option: `bundler: 'webpack' | 'esbuild' | false` (`false` would be the same as `webpack: false`, or a potential `esbuild: false`). In theory this can then also support other bundlers/plugins if there would be any in the future.

```js
// .size-limit.js

module.exports = [
  {
    name: 'first-package',
    bundler: 'esbuild',
  },
  {
    name: 'second-package',
    bundler: 'webpack',
  },
];
```

It'd be even crazier if people could use their own bundler if they want to:

```js
// .size-limit.js

const esbuildPlugin = require('@size-limit/esbuild');
const theirOwnPlugin = require('./size-limit-rolldown');

module.exports = [
  {
    name: 'first-package',
    bundler: esbuildPlugin,
  },
  {
    name: 'second-package',
    bundler: theirOwnPlugin,
  },
];
```

JPeer264:
closes #396 

This adds the option `disablePlugins`. The benefit of doing this is to still have multiple plugins installed, but disable plugins per entry of the array. The usage could look like this:

```js

modue.exports = [
  // following disables the `esbuild` plugin
  {
    name: 'first-package',
    disablePlugins: ['esbuild'],
  },
  // following disables the `webpack` plugin
  {
    name: 'second-package',
    disablePlugins: ['webpack'],
  },
];
```

I do wonder now if it would make sense to deprecate the `webpack: boolean` option in favor of `disablePlugins`, as `disablePlugins: ['webpack']` would do the same now

### DIFF
--- a/packages/size-limit/index.d.ts
+++ b/packages/size-limit/index.d.ts
@@ -20,11 +20,10 @@ export interface Check {
   disableModuleConcatenation?: boolean
 
   /**
-   * Array of plugin names to skip for this check.
-   * Plugin names can be short (`"esbuild"`), scoped (`"@size-limit/esbuild"`),
-   * or hyphenated (`"size-limit-esbuild"`).
+   * Plugin npm package names to skip for this check.
+   * For example: `["@size-limit/webpack"]`, `["@size-limit/esbuild"]`, or `["@size-limit/time"]`.
    */
-  disablePlugins?: string | string[]
+  disablePlugins?: string[]
 
   /**
    * When using a custom webpack config, a webpack entry could be given.
--- a/packages/size-limit/index.d.ts
+++ b/packages/size-limit/index.d.ts
@@ -20,6 +20,13 @@ export interface Check {
   disableModuleConcatenation?: boolean
 
   /**
+   * Array of plugin names to skip for this check.
+   * Plugin names can be short (`"esbuild"`), scoped (`"@size-limit/esbuild"`),
+   * or hyphenated (`"size-limit-esbuild"`).
+   */
+  disablePlugins?: string | string[]
+
+  /**
    * When using a custom webpack config, a webpack entry could be given.
    * It could be a string or an array of strings. By default,
    * the total size of all entry points will be checked.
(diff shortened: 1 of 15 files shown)


packages/size-limit/index.d.ts
```diff
@@ -19,6 +19,13 @@ export interface Check {
 
   disableModuleConcatenation?: boolean
 
+  /**
+   * Array of plugin names to skip for this check.
+   * Plugin names can be short (`"esbuild"`), scoped (`"@size-limit/esbuild"`),
```
ai: It looks too complicated.

For me, it feels like that it will be safer (for future maintenance) to just receive npm package name here. It will save us from name conflicts and make system more straight forward and simple.

Do you have any reasons for this more starter way?

JPeer264: This makes sense. I mainly took it, because `load-plugins` had a similar logic like this. But I do agree that having one approach is best. I'll update it to only accept the package name itself.

JPeer264: Alright, I just updated it to only accept the NPM package names directly. Furthermore I removed the ability to pass `string | string[]` and focused on only arrays instead.

ai:
Thanks. Released in 12.1.0.

PR #398, opened 2026-09-12
Add rolldown and rolldown-why plugins

dangreen:
Adds two packages that mirror the existing bundler adapters:

- `@size-limit/rolldown` builds the project with [Rolldown](https://rolldown.rs/),
  the way `@size-limit/esbuild` does with esbuild.
- `@size-limit/rolldown-why` shows what the bundle is made of,
  like `@size-limit/esbuild-why`.

Both follow the existing packages: same plugin steps, same `check` field naming,
same `package.json` shape, same test and fixture layout.

## Motivation

Rolldown's dead code elimination is more aggressive and more modern than
esbuild's. It supports `/*#__NO_SIDE_EFFECTS__*/`, which esbuild ignores — and
that is the annotation a library author can actually add, since it goes on the
declaration. esbuild only honours `/*#__PURE__*/`, which has to be written at
every call site, and call sites belong to the consumer.


ai:
Do you have some examples of the size diff vs esbuild or webpack?

dangreen:
@ai 

| import | esbuild | webpack (library) | rolldown |
| --- | --- | --- | --- |
| `agera` → `{ signal }` | 1344 B / 12 ms | 1268 B / 272 ms | 1303 B / 9 ms |
| `agera` → `{ signal, computed }` | 1429 B / 12 ms | 1352 B / 287 ms | 1379 B / 5 ms |
| `agera` → `{ signal, computed, effect }` | 1519 B / 11 ms | 1441 B / 276 ms | 1474 B / 6 ms |
| `agera` → `{ mountable }` | 481 B / 10 ms | 425 B / 247 ms | 401 B / 4 ms |
| `agera` → `{ selector }` | 1380 B / 11 ms | 1349 B / 255 ms | 1380 B / 4 ms |
| `kida` → `{ record }` | 1660 B / 16 ms | 1604 B / 335 ms | 1609 B / 9 ms |
| **total** | **72 ms** | **1673 ms** | **38 ms** |

Sizes: brotli. Times: fastest of 3 runs after a warm-up build.

ai:
Thanks. Released in 13.1.1.

PR #401, opened 2026-09-22
docs: fix copyable JSON and TypeScript config examples

Likio3000:
The Limits Config section shows JavaScript syntax under the `.size-limit.json` filename, so copying that block produces invalid JSON. Its TypeScript example also imports `SizeLimitConfig` from a repository-relative fixture path that is unavailable in an installed project.

Use JSON syntax and a `json` fence for the JSON example, and import the type from `size-limit`. The example values and module export style stay the same.

Validation: the original JSON example fails JSON parsing; the corrected block parses to the expected configuration array. Statically checked the config search paths, existing config fixtures/tests, and the package's exported `SizeLimitConfig` declaration. No Size Limit loader, bundle analysis, test suite, or TypeScript compiler was run.

AI disclosure: prepared with OpenAI Codex on behalf of Likio3000, with independent review before submission.


ai:
Thanks!

